Issue #9, opened 2018-07-31
Enhance "first" integration test to actually start server and check endpoint somehow.


scottkurz:
Currently it just verifies that we can do a mvn build successfully.

PR #14, opened 2018-07-31
Add endpoint ping to first IT

scottkurz:
Fixes issue #9


src/it/first/src/test/java/it/EndpointIT.java
```diff
@@ -0,0 +1,51 @@
+/*******************************************************************************
+ * (c) Copyright IBM Corporation 2017.
```
cherylking: Change copyright year to 2018 and reference EPL 1.0 license.

Issue #12, opened 2018-07-31
Add Apache 2.0 licenses to various files


cherylking:
Need to add the Apache 2.0 license to repository files and associate the license to the repo.

scottkurz:
One idea I started on was [this](https://github.com/scottkurz/boost/commit/8b1546076115e29a799caa4cf5855b3e7c323fb6) using maven-rat-plugin.

. It's only doing one line of matching (so it'd pass if you had just the one magic line but NONE of the rest). 

Maybe the https://www.mojohaus.org/license-maven-plugin/check-file-header-mojo.html is another option.

Or maybe it's not even worth it.

scottkurz:
Seems like the new home is:  http://code.mycila.com/license-maven-plugin/

PR #15, opened 2018-07-31
Detect spring boot dependency version

### LINKED ISSUE #5: Support SpringBoot 2.0 as an additional choice (vs. SpringBoot 1.5)
* Key off of parent starter POM version
* generate different Liberty server.xml features.

### LINKED ISSUE #9: Enhance "first" integration test to actually start server and check endpoint somehow.

cherylking:
Fixes #5  and part of #12

### DIFF
--- a/src/main/java/boost/project/PackageSpringBootAppWithLibertyMojo.java
+++ b/src/main/java/boost/project/PackageSpringBootAppWithLibertyMojo.java
@@ -13,7 +13,9 @@ package boost.project;
 import java.io.File;
 import java.net.URL;
 import java.util.List;
+import java.util.Set;
 
+import org.apache.maven.model.DependencyManagement;
 import org.apache.maven.model.Dependency;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecutionException;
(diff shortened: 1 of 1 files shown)


src/main/java/boost/project/PackageSpringBootAppWithLibertyMojo.java
```diff
@@ -223,4 +253,19 @@ private void createUberJar() throws MojoExecutionException {
...
 	}
+
+	/**
+	* Detect spring boot version dependency
+	*/
+	private String findSpringBootVersion(MavenProject project) {
+        	String version = null;
+        	List<Dependency> dependencies = project.getDependencies();
+        	for(Dependency dep: dependencies) {
+            		if("org.springframework.boot".equals(dep.getGroupId())){
```
scottkurz: Would it be better to narrow this down to the "spring-boot-starter-parent" artifact specifically within the group?   Not sure if there's a valid case where the parent starter comes from one version but some other dependency in the group comes from another (1.5.x vs. 2.0.x).

cherylking: This is how Anjum recommended checking the version of the spring boot dependency. The dependency versions end up coming from the parent pom (several levels of indirection up). Plus, the application could reference a different parent pom that then lists the spring boot one as its parent. In which case, the version would not be the one we are looking for.

cherylking: Let's talk more tomorrow about this mix version case and whether it is valid for Open Liberty.

cherylking:
@scottkurz Ready for another look.

scottkurz:
Reviewed.  Added https://github.com/OpenLiberty/boost/issues/21 for follow-up integration testing.

PR #20, opened 2018-08-01
Add EPL to code

scottkurz:


scottkurz:
Partial fix to issue #12. 

Since we've already moved past this.. I think we might as well move on, merge the PR, and  and leave the issue #12 open to consider how to verify this on an ongoing basis.

PR #25, opened 2018-08-03
Add pom.xml for integration tests

### LINKED ISSUE #24: Add a pom.xml to src/it to list all the integration tests as modules
We can also add some common properties and configuration as needed, instead of repeating it in each test pom.xml.

cherylking:
Fixes #24 

Also fixed a memory leak in the tests, and gave each application a unique name in their pom.xml files.


scottkurz:
Cancelling.  Will proceed with PR #27 (see comment there).

PR #27, opened 2018-08-03
Fixes #24; add IT-level POM, synch artifact and subdir names

### LINKED ISSUE #24: Add a pom.xml to src/it to list all the integration tests as modules

scottkurz:
Going to replace PR #25 with this, since this merges in an additional change, making the artifact name the same as the subdir names:  **test15Explicit**, **test15Implicit**, etc.


scottkurz:
Merging without review.  This is basically my review of the original PR #25 of @cherylking .

PR #36, opened 2018-08-07
Clean up project, use Liberty Maven Plugin 2.5.1-SNAPSHOT

### LINKED ISSUE #31: Don't run ITs by default
Activate using a profile instead. Can do this with #18.

chyt:
Closes #31.


pom.xml
```diff
@@ -1,79 +1,80 @@
...
-              </execution>
-          </executions>
-      </plugin>
-    </plugins>
-  </build>
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+        <maven.compiler.source>1.7</maven.compiler.source>
+        <maven.compiler.target>1.7</maven.compiler.target>
```
scottkurz: Thinking 1.8 is better.

src/it/pom.xml
```diff
@@ -1,25 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
scottkurz: Would like to keep this file.

Issue #40, opened 2018-08-08
Get liberty-maven-plugin version used by mojo-executor from pom.xml


scottkurz:
Instead of hard-coding in Java, it's probably easier to manage this project by externalizing this into the POM and then pass it along into the mojo-executor invocation.

This way people new to the project can more easily see what the dependency version is, for one.

scottkurz:
The most basic solution for #40 would be just having a hard-coded liberty plugin version in the boost POM. ... and the next level of configurability would be allowing it to be configurable via the user's pom.xml. 

That could either be a separate issue or perhaps even done along with this one.

scottkurz:
From Gitter  _for #40 (PR #44), I think I was letting the desire for "what version of ci.maven does boost default to?"  to be documented confuse my thinking here.   It's just something that should be documented ... and confusing the boost POM by adding this in some hack way isn't going to make this more comprehensible.   I know @chyt had the reaction, too, that this didn't make sense.   OK, I was wrong.... it looks good..just squash and test (when possible) and that can resolve #40._

PR #44, opened 2018-08-14
40 liberty maven plugin config

uberskigeek:
Updates to allow the liberty maven plugin version used by the boost plugin to be configured. Updated the test15Explicit to configure the plugin other test cases use the default value.


scottkurz:
@uberskigeek this is valuable in and of itself but doesn't address the point about the default liberty-maven-plugin version being buried in the Java rather than surfaced to the boost plugin POM.   Which was the original goal of #40.

I'd like to learn the pros and cons of doing this.

Also, probably worth squashing these commits.

PR #62, opened 2018-08-17
Fix issue #26 and #16

### LINKED ISSUE #16: Switch to final groupId:  io.openliberty.boost?
Now using "<groupId>boost.project</groupId>"

### LINKED ISSUE #26: Repackaging twice without changing application classes results in issues.
To recreate: 

1.  Run a top-level:
> mvn clean install -Pit
2.  cd to IT proj dir 
> cd target/it/test15Implicit
3. Do another build:
> mvn install    # or even `mvn liberty:start-server`

The **messages.log** shows:
`
 [8/3/18 13:02:31:608 EDT] 00000029 com.ibm.ws.app.manager.AppMessageHelper E CWWKZ0002E: An exception occurred while starting the application thin-gs-spring-boot-0.1.0. The exception message was: java.lang.RuntimeException: Failed to run jar due to error C:\Users\ScottKurz\wlpExtract\thin-gs-spring-boot-0.1.0_73998872913722\wlp\bin\tools\ws-server.jar (The system cannot find the path specified)`

chyt:
Closes #26 and closes #16.


chyt:
The name of the artifact is independent of whether or not it is the one installed. If you look at the PackageServerMojo, when the attach config param is set to true, we set the project artifact as the packages server. If you look at the output from the install phase in the tests, you will see that the LibertyThinPackage.jar is actually what is being installed.

boost-maven-plugin/src/main/java/boost/project/LibertyBoostMojo.java
```diff
@@ -91,16 +102,33 @@ public void execute() throws MojoExecutionException {
...
             throw new MojoExecutionException("Unable to generate server configuration for the Liberty server", e);
         }
 
         installMissingFeatures();
+        
+        // Copy the Spring Boot uber JAR back as the project artifact
+        try {
+            File springJar = new File(springBootUtil.getSpringBootUberJarPath(project.getArtifact().getFile()));
+            if(springJar.exists()) {
+                FileUtils.copyFile(springJar, project.getArtifact().getFile());
```
cherylking: As we discussed, this copy is not always necessary. If we just copied the SB uber JAR over at line 87, we do not need to copy it back. That utility method would have to return a boolean indicating that the copy occurred though.

boost-maven-plugin/src/main/java/boost/project/LibertyBoostMojo.java
```diff
@@ -292,21 +312,22 @@ private void installMissingFeatures() throws MojoExecutionException {
...
     private void createUberJar() throws MojoExecutionException {
-
         // Package server into runnable jar
         executeMojo(getPlugin(),
-                goal("package-server"), configuration(element(name("isInstall"), "false"),
-                        element(name("include"), "minify,runnable"), element(name("serverName"), libertyServerName)),
-                getExecutionEnvironment());
+            goal("package-server"), configuration(element(name("isInstall"), "false"),
+                    element(name("include"), "minify,runnable"),
+                    element(name("attach"), "true"),
```
cherylking: Do we still want "attach" specified here since we are not changing the name of the deployed archive?

chyt: Yes, otherwise the project artifact will be the thinned artifact (it is changed to that in the `install-apps` goal).

boost-maven-plugin/src/main/java/boost/project/LibertyBoostMojo.java
chyt: Opened https://github.com/OpenLiberty/boost-common/pull/8 to fix this.

boost-maven-plugin/src/main/java/io/openliberty/boost/LibertyBoostMojo.java
```diff
@@ -231,29 +256,16 @@ private ExecutionEnvironment getExecutionEnvironment() {
...
 		
-    	List<String> listOfDependencies = new ArrayList<String>();
-		getLog().debug("first lets see what dependencies we find");
-		for (Artifact artifact : mavenProject.getArtifacts()) {
-			getLog().debug("found this dependency, adding as a string -> " + artifact.getArtifactId());
-			listOfDependencies.add(artifact.getArtifactId());
-		}
-    	
-    	return listOfDependencies;
+		return featureStrings;
```
cherylking: Did you mean to remove the findJ2EEAppFeatureDependencies() method?

chyt: The method wasn't being used. Not sure if it was necessary.

PR #75, opened 2018-09-04
Closes #64

### LINKED ISSUE #64: Introduce boost run, debug, start (w/ optional debug), stop goals
Not much detail, I know.. this could be recast as an epic, separated out into four or more issues.

Should be similar wrapping of liberty-maven-plugin whether it is a SpringBoot or EE app.

chyt:
Closes #64.


chyt:
Fair point - opened https://github.com/OpenLiberty/boost-maven/issues/76

PR #82, opened 2018-09-07
Cleanup, test on Java 9

chyt:


scottkurz:
I don't think we should remove the EE 7 API dependencies along with the Java 9 move.. but we're discussing on Gitter now.

chyt:
The Java EE API dependency was removed in the test because it is already provided by the JAXRS feature dependency.

PR #94, opened 2018-09-10
Refactor Spring Boot utils to boost-common

### LINKED ISSUE #69: Error handling for boost:package goal
The question came up whether boost:package should be able to be called directly, given that the SB maven plugin package goal is required to be called first. We should determine if we are in the SB scenario, and if so check that the Uber jar is either the SB Uber jar or the Liberty Boost Uber jar before attempting to package. If it is not, an error should be thrown (or would we want to call the SB maven plugin package goal ourselves in that case?).

chyt:
Depends on merging of https://github.com/OpenLiberty/boost-common/pull/11. Fixes #69.


boost-maven-plugin/src/main/java/io/openliberty/boost/utils/MavenProjectUtil.java
```diff
@@ -0,0 +1,28 @@
+package io.openliberty.boost.utils;
```
cherylking: Needs copyright and license.

boost-maven-plugin/src/main/java/io/openliberty/boost/utils/BoostLogger.java
```diff
@@ -0,0 +1,53 @@
+package io.openliberty.boost.utils;
```
cherylking: Needs copyright and license.

PR #96, opened 2018-09-10
Fix docker documentation

anjumfatima90:


Tutorial.md
```diff
@@ -78,6 +78,10 @@ Edit the Boost plugin definition in the pom.xml to the following:
 
 Now, run `mvn clean install` to build your docker image.
 
+######Note:
```
barecode: Missing a space before Note, so it renders wrong.

anjumfatima90: fixed.

PR #101, opened 2018-09-13
Remove classifier variable and modify Dockerize constructor

anjumfatima90:


boost-maven-plugin/src/main/java/io/openliberty/boost/docker/AbstractDockerMojo.java
```diff
@@ -71,7 +62,7 @@
     /**
      * Connect to Docker Daemon using HTTP proxy, if set.
      */
-    @Parameter(defaultValue = "true", property = "useProxy")
```
chyt: You can remove `property = "useProxy"` since the variable is already named that.

scottkurz: Doesn't removing the 'property' attribute of the annotation prevent you from using a -D override from the command line?   E.g. `mvn... -DuseProxy=false`

chyt: You're right, didn't realize that. Ok, we can keep it then.

PR #103, opened 2018-09-14
Add tests for server run

### LINKED ISSUE #76: Test start/stop/debug/run goals

awisniew90:
Fixes #76 "run" scenario.

The "start" and "stop" scenarios are implicitly tested when each integration test calls start/stop before and after the tests are run. Any failures in those goals would be caught there. 

The "run" goal is tested by calling `mvn boost:run` on a separate thread in the tests "BeforeClass" method. This is instead of calling "start" and "stop" before and after the integration test phase. The test then checks that the application is reachable.

### DIFF
--- a/boost-maven-plugin/src/it/pom.xml
+++ b/boost-maven-plugin/src/it/pom.xml
@@ -21,6 +21,7 @@
         <module>test20Implicit</module>
         <module>testWebsocket</module>
         <module>test-docker</module>
+        <module>test-run-server</module>
     </modules>
     
 </project>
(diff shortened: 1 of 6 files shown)


scottkurz:
Agree we have some level of coverage of the start/stop Mojos with existing ITs binding them to IT phases.   Do we have any big gaps we should fill?

Anything unique to the default-cli execution or non-default config parms?

One thing that comes to mind might be the timeouts from the LMP which boost wraps.  

Would it be worth having a test with a quick timeout of 1 second that we expect to hit all or most of the time?

Hmm,  looking [here](https://github.com/WASdev/ci.maven/blob/master/docs/start-server.md) I wonder if it even makes sense for boost to expose a **verifyTimeout** since we don't use *application* do we?   Maybe I'm suggesting just testing **serverStartTimeout**.



Any thought on the debug goal/mojo?

PR #106, opened 2018-09-18
2 Enforce coming .java and .xml code format

uberskigeek:


pre-commit.template
```diff
@@ -0,0 +1,89 @@
...
+# set this to the location of your Eclipse installtion
+# Mac OSX Example: /Users/mac/EclipseInstalls/EclipseO2/EClipse.app/Contents/MacOS
+ECLIPSE_HOME=**********Fully qualified location of Eclipse executable ******
+
+# Style file is provided in the root directory of the project as java.perfs
+# The project location is the location of the boost-maven project on your
+# workstation. 
+# example: /Users/mac/OpenLiberty-boost-maven/boost-maven
+PROJECT_LOCATION=********Fully qualified location of this projects .git folder *****
+STYLE_FILE=$PROJECT_LOCATION/java.perfs
```
chyt: Should this be named `java.prefs`?

uberskigeek: Open to suggestions. how does javaFormat.perfs sound?

chyt: Oh, I just meant that the extension should be `prefs` (preferences) instead of `perfs`. Probably just a typo :)

uberskigeek: Ah!! My dyslexia strikes again!. I'll update.

pre-commit.template
```diff
@@ -0,0 +1,123 @@
...
+# This hook will run the eclipse code formatter before any commit.
+# This will format Java code files that are part of the commit.
+# The java code will be formatted based on the preferences defined
+# in the java.prefs file provided in the base directory of this project.
+#
+# Update the following items below then save this file in 
+# .git/hooks/pre-commit 
+# be sure you have execute authority for this file. 
+
+# set this to the location of your Eclipse installtion
```
cherylking: install**a**tion

pre-commit.template
```diff
@@ -0,0 +1,123 @@
...
+# Mac OSX Example: /Users/mac/EclipseInstalls/EclipseO2/EClipse.app/Contents/MacOS
```
cherylking: Might be nice to have an example for Windows and Linux too.

uberskigeek: Good Point, which has me thinking. This will obviously only work on linux/Mac/cygwin. Wondering if I should indicate this won't run in windows unless a unix environment is provided
(Cygwin) Which begs the question do we need a windows only version of this functionality?

PR #107, opened 2018-09-18
Fixes #105

### LINKED ISSUE #105: Preserve original JAR when doing Boost packaging of a SpringBoot app using repackage classifier
As noted [here](https://docs.spring.io/spring-boot/docs/current/maven-plugin/examples/repackage-classifier.html), the point is to preserve the orig. JAR so it can be used as a compile (or I guess other) dependency.     

So we should end up with:
* my.jar (original JAR)
* my-exec.spring (SB Uber JAR)
* my-exec.jar (Liberty Uber JAR)

But instead now we're getting:
*    my.jar (Liberty Uber JAR)
 *   my-exec.jar (SB Uber JAR)

Not sure how easy it is to have the test confirm that the original JAR is also "attached" (installed to Maven repo).  If it's not too hard might  that be worth doing in IT?

chyt:
Fixes #105. Depends on https://github.com/OpenLiberty/boost-common/pull/13.

### DIFF
--- a/boost-maven-plugin/src/main/java/io/openliberty/boost/liberty/LibertyPackageMojo.java
+++ b/boost-maven-plugin/src/main/java/io/openliberty/boost/liberty/LibertyPackageMojo.java
@@ -65,46 +65,83 @@ public class LibertyPackageMojo extends AbstractLibertyMojo {
         createDefaultRuntimeArtifactIfNeeded();
 
         createServer();
-        
+
         /**
          * Whether the packaged Liberty Uber JAR will be the project artifact. This
          * should be the case in Spring Boot scenarios since Spring Boot developers
-         * expect a runnable JAR. 
+         * expect a runnable JAR.
          */
         boolean attach;
 
+        /**
+         * Use the classifier to determine whether we need to set the Liberty Uber JAR
+         * as the project artifact, and add Spring-Boot-Version to the manifest
+         */
+        String springBootClassifier = null;
+
         try {
-            if (springBootVersion != null && !springBootVersion.isEmpty()) {
-                // Dealing with a spring boot app
-                attach = true;
-                validateSpringBootUberJAR();
-                copySpringBootUberJar();
+            if (MavenProjectUtil.isValid(springBootVersion)) { // Dealing with a spring boot app
+                springBootClassifier = net.wasdev.wlp.maven.plugins.utils.SpringBootUtil
+                        .getSpringBootMavenPluginClassifier(project, getLog());
+
+                // Check if we need to attach based on the classifier configuration
+                if (MavenProjectUtil.isValid(springBootClassifier)) {
+                    attach = false;
+                } else {
+                    attach = true;
+                }
+
+                File springBootUberJar = net.wasdev.wlp.maven.plugins.utils.SpringBootUtil.getSpringBootUberJAR(project,
+                        getLog());
+                validateSpringBootUberJAR(springBootUberJar);
+                copySpringBootUberJar(springBootUberJar, attach); // Only copy back if we need to overwrite the project
+                                                                  // artifact
                 installApp("spring-boot-project");
                 generateServerXML();
                 generateBootstrapProps();
-            } else {
-                // Dealing with an EE based app
+                installMissingFeatures();
+                
+                if(springBootUberJar != null) {
+                    createUberJar(springBootUberJar.getAbsolutePath(), attach);
+                }
+                else {
+                    createUberJar(null, attach);
+                }
+                
+                if (!MavenProjectUtil.isValid(springBootClassifier)) {
+                    // If necessary, add the manifest to prevent Spring Boot from repackaging again
+                    addSpringBootVersionToManifest(springBootVersion);
+                }
+            } else { // Dealing with an EE based app
                 attach = false;
                 installApp("project");
                 boosterFeatures = getBoosterConfigsFromDependencies(project);
                 generateServerXMLJ2EE(boosterFeatures);
+                installMissingFeatures();
+                createUberJar(null, attach);
             }
         } catch (TransformerException | ParserConfigurationException e) {
-            throw new MojoExecutionException("Unable to generate server configuration for the Liberty server", e);
+            throw new MojoExecutionException("Unable to generate server configuration for the Liberty server.", e);
         }
+    }
 
-        installMissingFeatures();
-
-        createUberJar(attach);
-
-        if (springBootVersion != null) {
-            // Add the manifest to prevent Spring Boot from repackaging again
+    /**
+     * Manipulate the manifest so that Spring Boot will not attempt to repackage a
+     * Liberty Uber JAR.
+     * 
+     * @param springBootVersion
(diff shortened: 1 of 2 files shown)


boost-maven-plugin/src/main/java/io/openliberty/boost/liberty/LibertyPackageMojo.java
```diff
@@ -131,15 +166,25 @@ private void validateSpringBootUberJAR() throws MojoExecutionException {
...
+    private void copySpringBootUberJar(File springBootUberJar, boolean attach) throws MojoExecutionException {
         try {
-            if (!SpringBootUtil.copySpringBootUberJar(project.getArtifact().getFile(), BoostLogger.getInstance())) {
-                File springJar = new File(
-                        SpringBootUtil.getBoostedSpringBootUberJarPath(project.getArtifact().getFile()));
-                if (net.wasdev.wlp.common.plugins.util.SpringBootUtil.isSpringBootUberJar(springJar)) {
-                    getLog().debug("Copying back Spring Boot Uber JAR as project artifact.");
-                    FileUtils.copyFile(springJar, project.getArtifact().getFile());
+            File springBootUberJarCopy = null;
+            if (springBootUberJar != null) {
```
scottkurz: If this test fails (and it's null), it seems to me like we should throw a MojoFailureExc (or ExecutionExc).   If all we're doing is avoiding an NPE.. well, I'd rather get an NPE than unexpected behavior we never planned for.

boost-maven-plugin/src/main/java/io/openliberty/boost/liberty/LibertyPackageMojo.java
```diff
@@ -131,15 +166,25 @@ private void validateSpringBootUberJAR() throws MojoExecutionException {
...
+                springBootUberJarCopy = SpringBootUtil.copySpringBootUberJar(springBootUberJar,
+                        BoostLogger.getInstance());
+            }
+
+            if (springBootUberJarCopy == null) { // Copy didn't happen
+                if (attach) {
```
scottkurz: What scenario causes us to fall into this path?   The copy to xxx.spring failed, but we're going to copy back from xxx.spring to the Uber JAR?    I'm not following.

boost-maven-plugin/src/main/java/io/openliberty/boost/liberty/LibertyPackageMojo.java
```diff
@@ -65,46 +65,83 @@ public void execute() throws MojoExecutionException {
...
                 generateBootstrapProps();
-            } else {
-                // Dealing with an EE based app
+                installMissingFeatures();
+                
+                if(springBootUberJar != null) {
+                    createUberJar(springBootUberJar.getAbsolutePath(), attach);
+                }
+                else {
+                    createUberJar(null, attach);
```
scottkurz: Maybe this is a repeat, but how could this be null?

boost-maven-plugin/src/main/java/io/openliberty/boost/liberty/LibertyPackageMojo.java
```diff
@@ -65,46 +65,83 @@ public void execute() throws MojoExecutionException {
...
+         */
+        String springBootClassifier = null;
+
         try {
-            if (springBootVersion != null && !springBootVersion.isEmpty()) {
-                // Dealing with a spring boot app
-                attach = true;
-                validateSpringBootUberJAR();
-                copySpringBootUberJar();
+            if (MavenProjectUtil.isValid(springBootVersion)) { // Dealing with a spring boot app
```
scottkurz: probably unneccessary right?  (Not that I hugely care, just wondering).

boost-maven-plugin/src/main/java/io/openliberty/boost/liberty/LibertyPackageMojo.java
```diff
@@ -65,46 +65,83 @@ public void execute() throws MojoExecutionException {
...
+                if(springBootUberJar != null) {
+                    createUberJar(springBootUberJar.getAbsolutePath(), attach);
+                }
+                else {
+                    createUberJar(null, attach);
+                }
+                
+                if (!MavenProjectUtil.isValid(springBootClassifier)) {
+                    // If necessary, add the manifest to prevent Spring Boot from repackaging again
+                    addSpringBootVersionToManifest(springBootVersion);
```
scottkurz: Hmm... I don't really have a strong reason, but my guess would have been to just add the manifest entry anyway to the Liberty Uber JAR.    That is:   "any time Boost packages a Liberty UBer JAR from SpringBoot app you see this manifest entry" rather than only adding it because we have a reason based on the packaging sequence.  That kind of consistency strikes me as better.

boost-maven-plugin/src/main/java/io/openliberty/boost/utils/MavenProjectUtil.java
```diff
@@ -34,5 +34,9 @@ public static String findSpringBootVersion(MavenProject project) {
 
         return version;
     }
+    
+    public static boolean isValid(String s) {
```
cherylking: Using this method on various strings makes it look like you are validating it is a valid spring boot version, or spring boot classifier etc. I suggest renaming from isValid to isNonNullAndNonEmpty or something like that.

boost-maven-plugin/src/main/java/io/openliberty/boost/liberty/LibertyPackageMojo.java
chyt: While this is possible, it adds more technical complexity since we then need a way of identifying the Liberty Uber JAR in both the classifier and non-classifier scenarios. It's not as simple as just getting the project artifact, which is what we currently do. Plus there's no additional value to the consistency, since we have other methods of identifying a Liberty Uber JAR (even in Spring Boot thin scenarios).

boost-maven-plugin/src/main/java/io/openliberty/boost/liberty/LibertyPackageMojo.java
chyt: `springBootUberJar` being null doesn't necessarily indicate a failure. This is for the case when a user is running the SB build a second time without cleaning. In that case we would not want to copy, but we would want to copy back (lines 179 to 183) given they are not using the classifier (attach is true).

boost-maven-plugin/src/main/java/io/openliberty/boost/liberty/LibertyPackageMojo.java
chyt: As stated in the previous comment, this is for when the user is running the build a second time without cleaning (what the invoker `mvn verify` tests test for)

boost-maven-plugin/src/main/java/io/openliberty/boost/utils/MavenProjectUtil.java
chyt: Sure, this can be changed.

PR #111, opened 2018-09-20
Validate <repository> and <tag> parameters

anjumfatima90:
Throw an error with a proper error message if these parameters are not set with valid names.

### DIFF
--- a/boost-maven-plugin/src/main/java/io/openliberty/boost/docker/DockerPushMojo.java
+++ b/boost-maven-plugin/src/main/java/io/openliberty/boost/docker/DockerPushMojo.java
@@ -49,8 +49,11 @@ public class DockerPushMojo extends AbstractDockerMojo {
             Scanner in = new Scanner(System.in);
             String newRepository = in.next();
             if (newRepository == null) {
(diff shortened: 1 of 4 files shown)


boost-maven-plugin/src/main/java/io/openliberty/boost/docker/DockerPushMojo.java
```diff
@@ -49,8 +49,11 @@ private void pushDockerImage(DockerClient dockerClient) throws MojoExecutionExce
             Scanner in = new Scanner(System.in);
             String newRepository = in.next();
             if (newRepository == null) {
-                throw new NullPointerException("Repository name cannot be null");
+                throw new MojoExecutionException("The repository name cannot be null");
             }
+            if (!isRepositoryValid(newRepository)){
+                throw new MojoExecutionException("The repository name is invalid.");
```
cherylking: should say "not valid" instead of "invalid"

anjumfatima90: done.

PR #121, opened 2018-09-24
Read server.address prop

scottkurz:


scottkurz:
This supersedes PR #119 (which didn't account for the need to change the part in common).

PR #143, opened 2018-10-10
Fix pre-commit template in case when there is a mix of staged/unstage…

scottkurz:
…d modifications


scottkurz:
Seems like a lot of hits say the Travis RC 137 means running out of memory or some host resource.  Retrying.

Issue #133, opened 2018-10-09
Boost docker-build goal doesn't work with classifier or as a standalone goal `mvn boost:docker-build` (without the 'jar' plugin)


scottkurz:
Details to follow..already started work on the fix...which will include a fix to LMP

chyt:
Details for this issue are described here: https://github.com/OpenLiberty/boost-maven/pull/144#pullrequestreview-163898934

PR #144, opened 2018-10-11
133 fix docker build with classifier

scottkurz:
Fixes issue #133

### DIFF
--- a/boost-maven-plugin/pom.xml
+++ b/boost-maven-plugin/pom.xml
@@ -34,7 +34,7 @@
         <dependency>
             <groupId>net.wasdev.wlp.maven.plugins</groupId>
             <artifactId>liberty-maven-plugin</artifactId>
-            <version>2.6.1</version>
+            <version>2.6.2-SNAPSHOT</version>
         </dependency>
 
         <!-- Other dependencies -->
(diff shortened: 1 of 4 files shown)


scottkurz:
@chyt  - The code as it's written in this PR fixes the problem where docker-build is invoked when there is a classifier.   It also fixes the case where docker-build is called after a previous package in an earlier build, e.g. `mvn boost:docker-build` .

I don't completely follow your comments.   It sounds like some overlap what I was trying to do in https://github.com/OpenLiberty/boost-common/pull/23, which is to print a more useful error message if you just tried to do `mvn boost:docker-build` without having done a package first.

If that's all your concern is (printing a better error message), can we just merge this PR and leave the error message discussion for https://github.com/OpenLiberty/boost-common/pull/23 ?

chyt:
Right, I want a more useful error message to be printed. However I think it should be done in `getAppArchive()` instead of in `boost-common`, since the helper methods are not specific to any scenario and thus should only provide more generic error messages. 

To restate my suggestion, I think you should sanity check the `project` object at the beginning of the `getAppArchive()` method, and if anything is wrong then throw an exception specifically related to the `docker-build` goal.

scottkurz:
@chyt  How about this approach (updated the PR)... put a more specific error here in boost-maven and we can forget the change in https://github.com/OpenLiberty/boost-common/pull/23.   (Note I left the boost-common dependency on 0.1.2-SNAPSHOT since we have new stuff there separately from this issue.

Also... since we've been back and forth on this exc message a few times.. (and I'm a bit concerned how much time I've spent on it)......if you want to suggest a modification on this how about doing so as a PR to my fork/branch?    

Thanks.

scottkurz:
Closes #133

PR #145, opened 2018-10-11
Docker IT cleanup and allow remote execution via DOCKER_HOST

scottkurz:


scottkurz:
I think adding support for DOCKER_HOST is probably not controversial.
(Note the repository name still uses 'localhost' in it even when the Docker engine is remote to the IT execution). 

As far as the refactoring of the stop/remove into `@AfterClass`, the problem addressed here is if the container creation succeeds but the test assertion fails, and the cleanup doesn't happen, then that affects the DockerPushIT tests and later execution of this tests.    So probably better to try to cleanup always.. though I'm sure there will be errors where people will have to deal with this manually in some from again.

PR #147, opened 2018-10-15
Add dynamic runtime and jdk tests

chyt:


.travis.yml
```diff
@@ -5,10 +5,26 @@ jdk:
     - openjdk8
     - oraclejdk8
     - oraclejdk9
-
+matrix:
+    exclude:
```
scottkurz: Why are these excludes here (sorry if you already explained)?

boost-maven-plugin/pom.xml
```diff
@@ -1,166 +1,202 @@
...
+								<goals>
+									<goal>install</goal>
+									<goal>run</goal>
+								</goals>
+							</execution>
+						</executions>
+					</plugin>
+				</plugins>
+			</build>
+		</profile>
```
scottkurz: Would be real nice if we could default from the cmd line so we didn't have to specify ` -DruntimeVersion=18.0.0.3 -Druntime=ol`.    I realize it's more work to add the logic, so could be a follow-up issue.

chyt: 18.0.0.3 is less compatible with Java 9 than 18.0.0.3, due to preparations to support Java 11 in 18.0.0.4. As a result, we don't test 18.0.0.3 on Java 9 in LMP. I think we should do the same thing here.

.travis.yml
chyt: 18.0.0.3 is less compatible with Java 9 than 18.0.0.2, due to changes in preparation for Java 11 support. As a result, we don't test 18.0.0.3 with Java 9 in LMP. I think we should do the same here.

boost-maven-plugin/pom.xml
chyt: This is added in 2a75aaa8b400ceda0e1aec9babffc86acbebf066.

Issue #17, opened 2018-08-01
Implicitly configure Liberty features based on Spring Boot configuration


awisniew90:
Any Liberty features that need to be defined in the server.xml should be implicitly determined based on the Spring Boot starter configuration.

scottkurz:
In particular:
* If the Spring boot application uses JSPs then a jsp feature is needed
* if WebSocket is needed then a webSocket feature is needed
* If https is required then the transportSecurity-1.0 feature is needed

Not sure if we want to take on calculating the version as part of this issue (jsp-2.2 or 2.3... websocket-1.0 or 1.1) or save until later.

awisniew90:
JSP configuration is dependent on war archive support for Spring Boot. (Typically, a Spring Boot project with JSPs would be packaged as a war). Moving that support over to issue #45

awisniew90:
The SSL support has been delivered to the Maven plugin.  This also needs to be ported to the Gradle plugin before this item can be closed.

PR #148, opened 2018-10-15
Add SSL support for Spring Boot path

awisniew90:
Adds SSL support for Issue #17 

Note:

1. The generateServerXML and generateBootstrap methods (both for spring boot) have been combined and  renamed to generateServerConfigSpringBoot. The LibertyServerConfigGenerator object in boost.common now handles the server.xml and bootstrap.properties files (entire config rather than just the xml file).

2. The execute method of the LibertyPackageMojo looks like a lot has changed, but the only changes are the calling of the combined method above, and the removal of the try/catch block. The mojo exceptions are now thrown directly from the generateServerConfigSpringBoot method.


boost-common/src/main/java/io/openliberty/boost/utils/SpringBootUtil.java
```diff
@@ -159,6 +172,33 @@ public static Properties getSpringBootServerProperties(String buildDir) throws I
...
+        String serverSslKeystoreProvider = (String) allProperties.getProperty(SERVER_SSL_KEYSTORE_PROVIDER);
+        String serverSslKeyAlias = (String) allProperties.getOrDefault(SERVER_SSL_KEY_ALIAS, "default");
+        String serverSslKeyPassword = (String) allProperties.getProperty(SERVER_SSL_KEY_PASSWORD);
+   
+        
+        if (serverSslKeystore != null && serverSslKeystorePassword != null ) {
+            serverProperties.setProperty(SERVER_SSL_KEYSTORE, serverSslKeystore);
+            serverProperties.setProperty(SERVER_SSL_KEYSTORE_PASSWORD, serverSslKeystorePassword);
+            
+            // Add any additional SSL properties
```
chyt: Why do you only add additional properties if the keystore and keystore password are null?

boost-common/src/main/java/io/openliberty/boost/utils/ConfigConstants.java
```diff
@@ -17,9 +17,21 @@
...
 
+    // KeyStore configuration values
+    public String KEYSTORE = "keyStore";
+    public String DEFAULT_KEYSTORE = "defaultKeyStore";
+    public String KEY_ENTRY = "keyEntry"; 
+    public String KEYSTORE_LOCATION = "location";
+    public String KEYSTORE_PASSWORD = "password";
+    public String KEYSTORE_TYPE = "type";
+    public String KEYSTORE_PROVIDER = "provider";
+    public String KEY_NAME = "name";
```
cherylking: I recommend removing KEY_ or KEYSTORE_ from any constant that is fairly generic and could be used on another config element in the future.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/liberty/LibertyPackageMojo.java
```diff
@@ -78,51 +84,71 @@ public void execute() throws MojoExecutionException {
...
+                attach = true;
+            }
 
+            File springBootUberJar = net.wasdev.wlp.maven.plugins.utils.SpringBootUtil.getSpringBootUberJAR(project,
+                    getLog());
+            validateSpringBootUberJAR(springBootUberJar);
+            copySpringBootUberJar(springBootUberJar, attach); // Only copy back if we need to overwrite the project
+                                                              // artifact
+
+            generateServerConfigSpringBoot();
```
chyt: The diff formatting is a bit weird in this method so it's hard to see what exactly was changed. So to clarify, the only thing that was changed here was to remove the try/catch and replace the `generateServerXML();` and `generateBootstrapProps();` methods with `generateServerConfigSpringBoot();`, is that correct?

boost-common/src/main/java/io/openliberty/boost/utils/SpringBootUtil.java
```diff
@@ -159,6 +172,33 @@ public static Properties getSpringBootServerProperties(String buildDir) throws I
         String serverAddress = (String) allProperties.getOrDefault(SERVER_ADDRESS, DEFAULT_SERVER_ADDRESS);
         serverProperties.setProperty(SERVER_PORT, serverPort);
         serverProperties.setProperty(SERVER_ADDRESS, serverAddress);
+        
+        String serverSslKeystore = (String) allProperties.getProperty(SERVER_SSL_KEYSTORE);
+        String serverSslKeystorePassword = (String) allProperties.getProperty(SERVER_SSL_KEYSTORE_PASSWORD);
+        String serverSslKeystoreType = (String) allProperties.getProperty(SERVER_SSL_KEYSTORE_TYPE);
```
cherylking: No reason to get the properties from 178 to 181 until the **if** stmt at line 184 is checked.

boost-common/src/main/java/io/openliberty/boost/utils/SpringBootUtil.java
```diff
@@ -206,4 +246,96 @@ public static Properties getSpringBootServerProperties(String buildDir) throws I
...
+        // Generate Liberty configuration
+        LibertyServerConfigGenerator serverConfig = new LibertyServerConfigGenerator(libertyServerPath);
+    
+        // Find and add appropriate springBoot features
+        List<String> featuresNeededForSpringBootApp = getLibertyFeaturesForSpringBoot(springBootVersion,
+                springBootStarters, logger);
+        
+        serverConfig.addFeatures(featuresNeededForSpringBootApp);
+         // Get Spring Boot server properties
+        Properties springBootServerProps = new Properties();
```
cherylking: This is not needed. The getSpringBootServerProperties method at line 276 creates and returns the Properties object. Just declare the variable type on that line.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/liberty/LibertyPackageMojo.java
```diff
@@ -78,51 +84,71 @@ public void execute() throws MojoExecutionException {
...
+     * @throws MojoExecutionException
+     */
+    private void generateServerConfigSpringBoot() throws MojoExecutionException {
+        
+        try {
+            // Get Spring Boot starters from Maven project
+            List<String> springBootStarters = MavenProjectUtil.getSpringBootStarters(project);
+            
+            // Set Liberty server path
+            String libertyServerPath = projectBuildDir + "/liberty/wlp/usr/servers/" + libertyServerName;
```
chyt: Can we set the install directory (`projectBuildDir + "/liberty/wlp"`) as a constant since we reference it in several places? It might change or be configurable in the future.

boost-common/src/main/java/io/openliberty/boost/utils/SpringBootUtil.java
```diff
@@ -206,4 +246,96 @@ public static Properties getSpringBootServerProperties(String buildDir) throws I
...
+            if (springBootServerProps.containsKey(SpringBootUtil.SERVER_SSL_KEYSTORE_PROVIDER)) {
+                keystoreProperties.put(ConfigConstants.KEYSTORE_PROVIDER, "${" + SpringBootUtil.SERVER_SSL_KEYSTORE_PROVIDER + "}");
+            }
+            
+            // Add any key properties to the separate key map.
+            if (springBootServerProps.containsKey(SpringBootUtil.SERVER_SSL_KEY_PASSWORD)) {
+                keyProperties.put(ConfigConstants.KEY_PASSWORD, "${" + SpringBootUtil.SERVER_SSL_KEY_PASSWORD + "}");
+            } 
+            if (springBootServerProps.containsKey(SpringBootUtil.SERVER_SSL_KEY_ALIAS)) {
+                keyProperties.put(ConfigConstants.KEY_NAME, "${" + SpringBootUtil.SERVER_SSL_KEY_ALIAS + "}");
```
cherylking: Since you are having to prepend `${` and append `}` for every one of these properties, why not declare constants with those already defined?

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/utils/MavenProjectUtil.java
```diff
@@ -37,5 +39,25 @@ public static String findSpringBootVersion(MavenProject project) {
...
         return s != null && !s.isEmpty();
     }
+    
+    /**
+     * Get all dependencies with "spring-boot-starter-*" as the artifactId. These
+     * dependencies will be used to determine which additional Liberty features need
+     * to be enabled.
+     * 
+     */
+    public static List<String> getSpringBootStarters(MavenProject project) {
```
chyt: Instead of getting the starters, I think we should look at the complete list of dependencies (including transitive) and check for the specific underlying Spring Framework dependencies. For example, instead of checking for `spring-boot-starter-web` we should check for `org.springframework:spring-webmvc` which is the underlying Spring MVC implementation. That way it is more reliable if the user has a Spring Boot app that is using Spring Framework dependencies, for whatever reason.

Also, for reliability I think we should be checking both the group ID and artifact ID, not just the artifact ID.

boost-common/src/main/java/io/openliberty/boost/utils/SpringBootUtil.java
chyt: Or, I was just thinking just have a helper method that takes in the string and wraps it in the brackets.

cherylking: But those values never change. So it really calls for constants.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/liberty/LibertyPackageMojo.java
cherylking: He covered this with his comment on the PR.

boost-common/src/main/java/io/openliberty/boost/utils/SpringBootUtil.java
awisniew90: We need at least the keystore (location) and password to configure our Liberty keystore element. I believe you also need at least this for spring boot. So if these are missing, we don't bother setting additional properties since the config is invalid.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/liberty/LibertyPackageMojo.java
awisniew90: Correct - for some reason, git butchered this (I think because the MojoException moved).

boost-common/src/main/java/io/openliberty/boost/utils/SpringBootUtil.java
```diff
@@ -159,6 +172,33 @@ public static Properties getSpringBootServerProperties(String buildDir) throws I
...
+                serverProperties.setProperty(SERVER_SSL_KEYSTORE_TYPE, serverSslKeystoreType);
+            }
+            if (serverSslKeystoreProvider != null) {
+                serverProperties.setProperty(SERVER_SSL_KEYSTORE_PROVIDER, serverSslKeystoreProvider);
+            }
+            
+            
+            if (serverSslKeyPassword != null) {
+                serverProperties.setProperty(SERVER_SSL_KEY_PASSWORD, serverSslKeyPassword);
+                serverProperties.setProperty(SERVER_SSL_KEY_ALIAS, serverSslKeyAlias);
```
cherylking: Why don't you check for the alias != null before setting this property? You do for all the others.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/liberty/LibertyPackageMojo.java
```diff
@@ -68,6 +70,8 @@ public void execute() throws MojoExecutionException {
         springBootVersion = MavenProjectUtil.findSpringBootVersion(project);
 
         boosterParent = new BoosterPacksParent();
+        
+        libertyServerPath = projectBuildDir + "/liberty/wlp/usr/servers/" + libertyServerName;
```
cherylking: I think Charles was suggesting that the quoted string here should be a constant.

boost-common/src/main/java/io/openliberty/boost/utils/SpringBootUtil.java
awisniew90: So from my basic understanding of how this works, our "keyEntry" element in server.xml requires a password, but does not necessarily need an alias. In that case, the "name" attribute of "keyEntry" is "default."  When we are retrieving the spring boot properties, if "server-ssl-key-alias" is not provided, we default it to "default" so it will never be null at this point. If there is a password, we create this "keyEntry" with password and name.  If password isnt present, we do nothing.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/liberty/LibertyPackageMojo.java
chyt: Not the whole quoted string - just the install directory (wlp folder) - which is `         libertyServerPath = projectBuildDir + "/liberty/wlp"`. Setting that as a constant and referencing file paths relative to the install directory variable would make the code more readable and flexible (in case the install directory changes or is configurable in the future).

awisniew90: hmm.  we only ever use that quoted string in reference to the full path to the server (i.e. projectBuildDir + string + libertyServerName).  I think it might be cleaner to have the full libertyServerPath variable defined to avoid that same string concatenation in multiple places.

chyt: Ok, that's fine. If we need to change it later we can.

PR #149, opened 2018-10-15
EE booster support: boost BOMs and generic boosters

ajm01:
liberty boost EE BOMs and generic booster support


scottkurz:
I did a rework here:
https://github.com/scottkurz/boost-maven/tree/ee-jaxrs-bom-4

Let's discuss in Gitter

scottkurz:
Some of the organization is a mix of "more art than science" and "too early to tell" (what the best groupings/folder hierarchies should be)... at least to me. But it moves the boosters to GAV like: io.openliberty.boosters:jaxrs:<version> and versions the BOMs along with boost-maven-plugin itself. Also (and we didn't talk about this) creates a new pair of BOMs: io.openliberty.boost:boost-ee7-apis-bom:0.1-SNAPSHOT and io.openliberty.boost:boost-ee8-apis-bom:0.1-SNAPSHOT. These manage the dependency versions both for user POMs which import the "external" Boost BOMs: io.openliberty.boosters:ee7-bom:<version> and io.openliberty.boosters:ee8-bom:<version> as well as for the booster POMs themselves. (Not convinced that's really helpful although it solves DNRY w.r.t. the jax-rs API version id).

PR #154, opened 2018-10-18
Allow repeat boost package; fixes #153 - DON'T MERGE

### LINKED ISSUE #153: For SB, boost:package goal doesn't work following 'package' phase: `mvn package;  mvn boost:package` doesn't work
I think we end up with a NPE, because, as in https://github.com/OpenLiberty/boost-maven/issues/133, the `project.getArtifact().getFile()` call returns 'null' if an earlier goal (e.g. jar:jar) hasn't been called.

Will include more detail in the PR I attempted for this, which didn't completely work.

scottkurz:
DON'T MERGE THIS !

Even if it passes CI, it fails some other tests.

### DIFF
--- a/boost-maven-plugin/src/it/test-spring-boot-1.5-explicit/invoker.properties
+++ b/boost-maven-plugin/src/it/test-spring-boot-1.5-explicit/invoker.properties
@@ -1,3 +1,5 @@
 invoker.goals.1 = clean install
 # purposely avoid clean second time
 invoker.goals.2 = verify
+# avoid clean, avoid earlier build phase
+invoker.goals.3 = boost:package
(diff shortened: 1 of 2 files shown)


scottkurz:
One note is there's an asymmetry here, in that the SBMP seems to rebuild the `xxx-<classifier>.jar` no matter what (at least in my experiments), whereas for the non-classifier case there's logic to look to see if it's already a SpringBoot Uber JAR, and if so, do nothing.

So we can prevent our Liberty Uber JAR from getting overwritten, in the non-classifier case, by putting a special entry in the JAR manifest.   But with the classifier case, this doesn't seem to be an option for us.

Even when I modified
```
                if (MavenProjectUtil.isNotNullOrEmpty(springBootClassifier)) {
                    // If necessary, add the manifest to prevent Spring Boot from repackaging again
                    addSpringBootVersionToManifest(springBootVersion);
                }
```
to do the add unconditionally, the end-to-end still didn't work.   (Don't remember if I left it in the PR or not).

scottkurz:
At a high level, it seems like "where I left off" in this PR, it seems like somehow we have to copy back from `xxx.jar.spring` back to `xxx.jar` so it can be deployed by the LMP **install-apps** goal.

I thought that might be accomplished by removing the `if(attach)` check, but that didnt' help.

scottkurz:
This isn't a high priority.  Closing.

PR #157, opened 2018-10-22
Migrate to a singular Boost repository

### LINKED ISSUE #152: Merge boost-maven, boost-common, boost-gradle
As discussed, this will make managing the backlog easier.

chyt:
Closes #152.


cherylking:
Should there be a readme at the top level?

scottkurz:
This could be a follow-up issue.. but what would you think about adding a top-level pom.xml, build.gradle, that built boost-common + boost-maven, and boost-common + boost-gradle, respectively?     Again, doesn't have to be done now, but thought it was worth mentioning.

chyt:
> Should there be a readme at the top level?

Yes, if someone can draft that, we can add that in a future PR. The old README was specific to Maven though so that's why it's in the `boost-maven` folder.

> This could be a follow-up issue.. but what would you think about adding a top-level pom.xml, build.gradle, that built boost-common + boost-maven, and boost-common + boost-gradle, respectively? Again, doesn't have to be done now, but thought it was worth mentioning.

Not sure if a pom.xml or build.gradle file is the best option, maybe a bash script instead? Either way I think it only makes the build marginally easier, so I wouldn't say it is high priority.

Issue #131, opened 2018-10-09
Build of docker image fails w/ open file handles, esp. on Windows - [INFO] Caused by: java.io.IOException: request to write '8192' bytes exceeds size in header ...


scottkurz:
As mentioned, this happens especially on Windows.

E.g. you can recreate running the integration test via `mvn install -Pit -Dinvoker.test=*2.0-implicit`

I imagine the fix involves `.dockerignore`...but I haven't looked to understand exactly what we're doing today.

I feel like this might not happen 100% of the time for me but I haven't proven that yet... it might be consistent.

Note the **build.log** in the exc is the **build.log** used to capture the log for the child, IT (maven-invoker) build.

```
[INFO] Caused by: org.apache.maven.plugin.MojoExecutionException: request to write '8192' bytes exceeds size in header of '6793' bytes for entry 'build.log'
[INFO]  at io.openliberty.boost.docker.DockerBuildMojo.execute(DockerBuildMojo.java:106)
[INFO]  at io.openliberty.boost.docker.AbstractDockerMojo.execute(AbstractDockerMojo.java:95)
[INFO]  at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:134)
[INFO]  at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:208)
[INFO]  ... 20 more
[INFO] Caused by: java.io.IOException: request to write '8192' bytes exceeds size in header of '6793' bytes for entry 'build.log'
[INFO]  at org.apache.commons.compress.archivers.tar.TarArchiveOutputStream.write(TarArchiveOutputStream.java:377)
[INFO]  at java.nio.file.Files.copy(Files.java:2909)
[INFO]  at java.nio.file.Files.copy(Files.java:3069)
[INFO]  at com.spotify.docker.client.CompressedDirectory$Visitor.visitFile(CompressedDirectory.java:303)
[INFO]  at com.spotify.docker.client.CompressedDirectory$Visitor.visitFile(CompressedDirectory.java:253)
[INFO]  at java.nio.file.Files.walkFileTree(Files.java:2670)
[INFO]  at com.spotify.docker.client.CompressedDirectory.create(CompressedDirectory.java:115)
[INFO]  at com.spotify.docker.client.DefaultDockerClient.build(DefaultDockerClient.java:1472)
[...]

PR #160, opened 2018-10-24
Issue131-Build of docker image fails w/ open file handles

anjumfatima90:
for issue #131


boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/docker/Dockerize.java
```diff
@@ -137,4 +141,55 @@ private void writeSpringBootDockerFile(File dockerFile, String libertyImage) thr
                 + "/config/dropins/spring/thin-${APP_FILE}");
         Files.write(dockerFile.toPath(), lines, Charset.forName("UTF-8"));
     }
+
+    public void createDockerIgnore() throws Exception {
```
scottkurz: maybe should be called `writeDockerIgnore()` or `mergeDockerIgnore()` right at the top-level here (instead of further down)?

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/docker/Dockerize.java
```diff
@@ -137,4 +141,55 @@ private void writeSpringBootDockerFile(File dockerFile, String libertyImage) thr
...
+        ArrayList<String> lines = new ArrayList<>();
+        if (pluginAlreadyExecuted(dockerIgnore)) {
+            return;
+        }
+
+        //Ignore the build.log getting generated by integration tests and the liberty runtime folder
+        if (dockerIgnoreAlreadyExists && dockerIgnore.length() > 0) {
+            lines.add("\n");
+        }
+        lines.add("# The following lines are added by boost-maven-plugin");
```
scottkurz: We could make this a constant so we don't have to repeat it in code.   It also should be common across gradle I'd think.    So maybe something more like  `...static final String BOOST_GEN = "# GENERATED BY LIBERTY BOOST PLUGIN"

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/docker/Dockerize.java
```diff
@@ -137,4 +141,55 @@ private void writeSpringBootDockerFile(File dockerFile, String libertyImage) thr
...
+        if (pluginAlreadyExecuted(dockerIgnore)) {
+            return;
+        }
+
+        //Ignore the build.log getting generated by integration tests and the liberty runtime folder
+        if (dockerIgnoreAlreadyExists && dockerIgnore.length() > 0) {
+            lines.add("\n");
+        }
+        lines.add("# The following lines are added by boost-maven-plugin");           
+        lines.add("build.log");
```
scottkurz: I think it'd be cleaner to add "build.log" in a `.dockerignore` associated with each IT using docker.  Not sure it's important enough to generate for everyone.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/docker/Dockerize.java
```diff
@@ -137,4 +141,55 @@ private void writeSpringBootDockerFile(File dockerFile, String libertyImage) thr
...
+
+    private void writeDockerIgnore(File dockerIgnore, boolean dockerIgnoreAlreadyExists) throws IOException {
+        ArrayList<String> lines = new ArrayList<>();
+        if (pluginAlreadyExecuted(dockerIgnore)) {
+            return;
+        }
+
+        //Ignore the build.log getting generated by integration tests and the liberty runtime folder
+        if (dockerIgnoreAlreadyExists && dockerIgnore.length() > 0) {
+            lines.add("\n");
```
scottkurz: Might be OK but just to be safe replace "\n" with `System.getProperty("line.separator")`

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/docker/Dockerize.java
```diff
@@ -137,4 +141,55 @@ private void writeSpringBootDockerFile(File dockerFile, String libertyImage) thr
...
+                while ((line = bufferedReader.readLine()) != null) {
+                   if ("# The following lines are added by boost-maven-plugin".equals(line)) {
+                       return true;
+                   }
+                }
+            }
+        }
+        return false;
+    }
 }
```
scottkurz: Two final, more general comments:

1. We should doc the behavior in the presence of existing .dockerignore(s) and also Dockerfile(s), at least in the form of a simple Javadoc comment in the mojo, which incorporates this into the help printed via  `mvn boost:help -Ddetail=true -Dgoal=docker-build`.      Just a simple one-line per behavior I think is a good start, all we need to do w/ this PR.

2. What to do about Gradle?  Assuming this doesn't get merged before the big Gradle code drop into this repo, we could just wait and keep going on this PR.   We could also open a separate issue.   Let's just do one or the other before closing this PR/issue.

scottkurz:
@anjumfatima90   - Instead of making comments, I did my own branch here:  https://github.com/scottkurz/boost-maven/tree/131-scott-code-review.  

I noticed you hadn't incorporated the [pre-commit](https://github.com/OpenLiberty/boost/blob/master/pre-commit.template) formatting hook, so all I did in my first commit is to squash your commits into one commit, with formatting added.   My 2nd commit just adds some text so that ` mvn boost:help -Ddetail=true -Dgoal=docker-build`  gives more detail about Dockerfile and .dockerignore generation.   Please review this text.

Will let you tell me if you want to update your PR with this or if you want to cancel this PR and have me start a new one.

scottkurz:
Replaced by PR #176

PR #164, opened 2018-10-25
Adding boost-gradle plugin

mattbsox:


boost-gradle/src/test/groovy/DockerBuild15Test.groovy
```diff
@@ -0,0 +1,119 @@
...
+        CreateContainerResponse container = dockerClient.createContainerCmd("localhost:5000/test-image15:latest")
+                .withPortBindings(PortBinding.parse("9080:9080")).exec()
+        Thread.sleep(3000)
+
+        dockerClient.startContainerCmd(container.getId()).exec()
+
+        Thread.sleep(10000)
+        testDockerContainerRunning()
+
+        Thread.sleep(3000)
```
scottkurz: I don't think this is long enough.. it failed for me about five times in a row on my Linux VM.     Or, I guess we could take it a step further and introduce some kind of polling

boost-gradle/src/test/groovy/DockerBuild15Test.groovy
```diff
@@ -0,0 +1,119 @@
...
+        testAppRunningOnEndpoint()
+
+        dockerClient.stopContainerCmd(container.getId()).exec()
```
scottkurz: The stop and remove should be in `@After` or  `@AfterClass`... when the testAppRunningOnEndpoint() fails, it seems to cause later tests to fail (not 100% sure this is what I saw but it seems a likely explanation)

Issue #139, opened 2018-10-09
Support Java EE 8


cherylking:
This epic covers all the items that need to be completed in order to say we fully support Java EE 8 applications in boost.

First the bom needs to be created that pulls in all the transitive dependencies. Then boosters need to be created for all the Java EE 8 technologies.

scottkurz:
We have basic support for Java EE 8 so marking this resolved.

Issue #138, opened 2018-10-09
Support Java EE 7


cherylking:
This epic covers all the items that need to be completed in order to say we fully support Java EE 7 applications in boost.

First the bom needs to be created that pulls in all the transitive dependencies. Then boosters need to be created for all the Java EE 7 technologies.

scottkurz:
We have basic support for EE 7 so marking this resolved.

Issue #135, opened 2018-10-09
Create a JAX-RS booster for Java EE 8


cherylking:
This booster should install the correct feature and configure it in the Liberty server.xml.

ajm01:
shipped in https://github.com/OpenLiberty/boost/pull/165

PR #165, opened 2018-10-25
New boosters and boms

ajm01:
This PR incorporates the original final proposal for generic boosters and BOMs which scott enhanced after the walk through last week. This work has also been incorporated into the new repo structure with this PR.

addresses issues:
https://github.com/OpenLiberty/boost/issues/134
https://github.com/OpenLiberty/boost/issues/135

first step for epics:
https://github.com/OpenLiberty/boost/issues/138
https://github.com/OpenLiberty/boost/issues/139


boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/BoosterPacksParent.java
```diff
@@ -32,8 +33,10 @@
         for (String dep : dependencies) {
             if (dep.equals(JDBC_BOOSTER_PACK_STRING)) {
                 boosterPackConfigList.add(new JDBCBoosterPackConfigurator());
-            } else if (dep.equals(JAXRS_BOOSTER_PACK_STRING)) {
-                boosterPackConfigList.add(new JAXRSBoosterPackConfigurator());
+            } else if (dep.equals(JAXRS_BOOSTER_PACK_STRING_10) || dep.equals(JAXRS_BOOSTER_PACK_STRING_20)) {
```
cherylking: Could we just check for startsWith("io.openliberty.boosters:jaxrs:") instead? Just thinking about having to change this when new releases of the boosters come out. Would this code really need to change?

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/JAXRSBoosterPackConfigurator.java
```diff
@@ -5,9 +5,12 @@
 
 public class JAXRSBoosterPackConfigurator implements BoosterPackConfigurator {
 
+	//default to the EE8 feature
+	String featureGAV = "jaxrs-2.1";
```
cherylking: Wondering if we really want to default here? What happens when the next Jakarta EE version comes out? Do we leave this as the default, or change to the latest? Are we always calling setFeatureString method regardless of the booster version?

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/BoosterPacksParent.java
```diff
@@ -15,7 +15,8 @@
 
     JDBCBoosterPackConfigurator jdbcConfig = null;
     private String JDBC_BOOSTER_PACK_STRING = "liberty-booster-data-jdbc";
-    private String JAXRS_BOOSTER_PACK_STRING = "io.openliberty.features:jaxrs-2.0";
+    public static String JAXRS_BOOSTER_PACK_STRING_10 = "io.openliberty.boosters:jaxrs:0.1-SNAPSHOT";
```
cherylking: General comment for all files....need copyright and license at top.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/JDBCBoosterPackConfigurator.java
```diff
@@ -9,6 +9,8 @@
 public class JDBCBoosterPackConfigurator implements BoosterPackConfigurator {
 
     private String jdbcDefault = "jdbc-4.1";
+    
+    String featureGAV = null;
```
cherylking: Should remove System.out.println() statements throughout.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/JDBCBoosterPackConfigurator.java
```diff
@@ -56,4 +58,10 @@ public void writeConfigToServerXML(Document doc) {
         lib.appendChild(fileLoc);
         serverRoot.appendChild(lib);
     }
+
+	@Override
+	public void setFeatureString(String feature) {
+		featureGAV = feature;
```
cherylking: Shouldn't getFeatureString() return featureGAV?

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/BoosterPacksParent.java
ajm01: i can do that - but I want to point out that this is the area of code that I think needs a general re-thought on design ideas. Basically was making this work for a prototype, but i don't think this approach is very efficient in the long run.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/JDBCBoosterPackConfigurator.java
ajm01: yup, i will hunt them down.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/JDBCBoosterPackConfigurator.java
ajm01: i conceived of this as a config generator that might render different config for different levels of the technology. not saying this is the right thing to do, but made it work that way - so once we know what level of technology we are dealing with via the dependency resolution, i am setting the specific value here so when the server.xml config step occurs this code will have a control point to write out the right config. having said that i am not utilizing a getter for that string ... so maybe i should use another approach for getting that technology string passed in.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/JAXRSBoosterPackConfigurator.java
ajm01: yes we are calling it everytime, so i could get rid of this default value

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/BoosterPacksParent.java
```diff
@@ -30,10 +41,14 @@
 
         featureList = new ArrayList<String>();
         for (String dep : dependencies) {
+        	System.out.println("AJM: dep = " + dep);
```
scottkurz: Still have one println()

boost-maven/boost-maven-plugin/src/it/test-jaxrs-2.0/pom.xml
```diff
@@ -31,22 +31,20 @@
...
 		<dependencies>
-			<dependency>
-				<groupId>io.openliberty.boost</groupId>
-				<artifactId>boost-bom</artifactId>
-				<version>@pom.version@</version>
-				<scope>import</scope>
-				<type>pom</type>
-			</dependency>
+		<dependency>
+			<groupId>io.openliberty.boosters</groupId>
```
scottkurz: There was a comment in an internal call or Gitter that this "groupId" should be "io.openliberty.boost.boosters" instead.   Personally, since I think this isn't the last input we'll hear on this, I think it's OK to merge either way, as long as we highlight this as an open question in ongoing design reviews.

boost-maven/boosters/booster-jaxrs-ee7/pom.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
scottkurz: The plugin pom.xml(s) at least should have copyrights.     Perhaps debatable if the IT poms are as big a deal (though easy enough to add them to all).

cherylking: I just checked ci.maven and we do not have a copyright in the pom.xml. So maybe it needs to be added in both places?

scottkurz: I'm not sure of a hard and fast rule.. it seems like a good place to draw the line to me, though.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/BoosterPacksParent.java
```diff
@@ -32,8 +43,11 @@
         for (String dep : dependencies) {
             if (dep.equals(JDBC_BOOSTER_PACK_STRING)) {
                 boosterPackConfigList.add(new JDBCBoosterPackConfigurator());
-            } else if (dep.equals(JAXRS_BOOSTER_PACK_STRING)) {
-                boosterPackConfigList.add(new JAXRSBoosterPackConfigurator());
+            //} else if (dep.equals(JAXRS_BOOSTER_PACK_STRING_10) || dep.equals(JAXRS_BOOSTER_PACK_STRING_20)) {
+            } else if (dep.startsWith("io.openliberty.boosters:jaxrs:")) {
```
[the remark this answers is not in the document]
ajm01: the JDBCBoosterPackConfig object was something i introduced many many weeks ago (at the outset of the EE booster pack effort) and it has carried along while all of my focus shifted to the JAXRS booster after we got some clarity on what we wanted to do with the booster packs. I ahve not put much effort into it per sey and really should have stripped it out until I get back to it with the jaxrs booster finalized in terms of the  booster support. 

That being said - the jaxrs 'else if' test is a response to Cheryl's comments earlier in the PR. Ultimately  I believe a more rigorous approach will be put in place for all of this config framework code - this is really prototype code designed to get this moving. I am removing the commented out code as i agree it is confusing leaving it in there...

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/liberty/LibertyPackageMojo.java
```diff
@@ -231,8 +231,8 @@ private void copySpringBootUberJar(File springBootUberJar, boolean attach) throw
 
         for (Artifact artifact : project.getArtifacts()) {
             getLog().debug("getBoostCfg: found this, adding as a string -> " + artifact.getGroupId() + ":"
```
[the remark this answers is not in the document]
ajm01: yes

boost-maven/boosters/booster-jaxrs-ee7/invoker.properties
```diff
@@ -0,0 +1 @@
+invoker.goals.1 = clean install
```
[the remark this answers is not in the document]
ajm01: ok, i locally removed these invoker prop files and the builds still do work as intended, so will delete them.

boost-maven/boosters/booster-jaxrs-ee8/invoker.properties
```diff
@@ -0,0 +1 @@
+invoker.goals.1 = clean install
```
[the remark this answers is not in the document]
ajm01: previous response applies here as well

PR #166, opened 2018-10-26
Move .gitignore back to top-level

scottkurz:


scottkurz:
Looking at Matt's PR https://github.com/OpenLiberty/boost/pull/164/files   I see that the Gradle subtree has a different .gitignore.  I'm going to rework to copy this over to common instead of putting in the root.

PR #167, opened 2018-10-26
Fix mvn path from IT on Windows

scottkurz:


scottkurz:
Seems from articles like this:
https://www.mkyong.com/java/how-to-detect-os-in-java-systemgetpropertyosname/

that this may be a specific-enough  test

scottkurz:
This works if `mvn` is in the PATH but I'd prefer to avoid worrying about my PATH.

boost-maven/boost-maven-plugin/src/it/test-run-server/src/test/java/it/LibertyRunMojoIT.java
```diff
@@ -22,30 +22,46 @@
...
 
+    private static String mvnCmd;
+
+    public static void setupMvnPath() {
+
+        String mvnPath = System.getenv("M2_HOME") + "/bin/mvn";
+
+        String osName = System.getProperty("os.name").toLowerCase();
+
+        if ((osName.indexOf("win") >= 0)) {
```
chyt: Here, please use the OSUtil method to check for Windows: https://github.com/WASdev/ci.common/blob/master/src/main/java/net/wasdev/wlp/common/plugins/util/OSUtil.java

boost-maven/boost-maven-plugin/src/it/test-run-server/src/test/java/it/LibertyRunMojoIT.java
```diff
@@ -22,30 +22,46 @@
...
+            mvnCmd = "cmd.exe /c " + mvnPath;
+        } else {
+            mvnCmd = mvnPath;
+        }
+    }
+
     @BeforeClass
     public static void init() throws Exception {
-        URL = "http://localhost:8080/";
```
chyt: This is just cleanup of an unused variable right?

chyt: Never mind, saw you moved it to the class level.

Issue #129, opened 2018-10-02
Use a boost standard context root for EE application.


scottkurz:

scottkurz:
Looks like by default you get the app=artifact name.... `Web application available (default_host): http://localhost:9080/test-jaxrs-2.0-1.0-SNAPSHOT/`   Maybe we should use `/boost`?

PR #169, opened 2018-10-31
Context root pr

ajm01:
this PR addresses https://github.com/OpenLiberty/boost/issues/129

### DIFF
--- a/boost-common/src/main/java/io/openliberty/boost/utils/LibertyServerConfigGenerator.java
+++ b/boost-common/src/main/java/io/openliberty/boost/utils/LibertyServerConfigGenerator.java
@@ -215,4 +215,13 @@ public class LibertyServerConfigGenerator {
               bootstrapProperties.put(key, value);
         }
     }
+    
+    public void addConfigForApp(String artifactId, String version){
+    	Element appCfg = doc.createElement(APPLICATION);
+    	appCfg.setAttribute("context-root", "/");
+    	appCfg.setAttribute("location", artifactId+"-"+version+".war");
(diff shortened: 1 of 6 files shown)


boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/utils/MavenProjectUtil.java
```diff
@@ -61,4 +61,8 @@ public static boolean isNotNullOrEmpty(String s) {
         return springFrameworkDependencies;
     }
 
+    public static String getArtictIdFromProject(MavenProject proj){
```
chyt: Don't think this is used anywhere, is it still needed?

boost-common/src/main/java/io/openliberty/boost/utils/LibertyServerConfigGenerator.java
```diff
@@ -215,4 +215,13 @@ public void addBootstrapProperties(Properties properties) throws IOException{
               bootstrapProperties.put(key, value);
         }
     }
+    
+    public void addConfigForApp(String artifactId, String version){
+    	Element appCfg = doc.createElement(APPLICATION);
+    	appCfg.setAttribute("context-root", "/");
+    	appCfg.setAttribute("location", artifactId+"-"+version+".war");
```
chyt: Is it safe to assume that we'll always be building a WAR? Should we get the packaging type dynamically instead?

ajm01:
closing and using #170 to ship both functions

PR #189, opened 2018-11-09
Refactor docker functionality to boost-common

chyt:


scottkurz:
So this PR will just focus on the maven plugin, moving its function to common, and we'll hook in gradle support later?   

Noting a few of the renamed/moved files now don't have copyrights.   Will try to look more...

chyt:
boost-maven and boost-common tests are passing. boost-gradle tests are failing due to a connection error for the endpoint tests. I'm proceeding with the merge so we can release boost-common and boost-maven, and will revisit boost-gradle tests tomorrow.

PR #190, opened 2018-11-12
Add a very hacky version of dockerizerJVM property

barecode:
For initial review. I want to propose another property. This does not need to be merged in right away.


chyt:
Not sure what exactly the `style` property is, can you provide some more details?

Also, can this be rebased on top of the latest master, and also be updated for boost-gradle?

PR #200, opened 2018-11-13
Expose ports in remaining docker tests

scottkurz:


scottkurz:
This is now captured in PR #203

PR #203, opened 2018-11-13
fixing gradle test issues and forwarding test log output

mattbsox:


scottkurz:
@mattbsox  - is d5d6e6701b4af662b0a6714ac9d03348c3734248  for the oracle JDK 9 noise?

mattbsox:
Ya, we thought that might be causing the warnings but it doesn't look like they're related. Still unnecessary for the tests though.

.travis.yml
```diff
@@ -37,8 +51,17 @@ before_install:
 install: true
 script:
     - cd boost-common
-    - travis_wait mvn install -Dinvoker.streamLogs=true
+    - mvn install -Dinvoker.streamLogs=true
     - cd ../boost-maven
-    - travis_wait mvn install -Pit -Dinvoker.streamLogs=true -Druntime=$RUNTIME -DruntimeVersion=$RUNTIME_VERSION
-    - cd ../boost-gradle
-    - travis_wait ./gradlew install check -Druntime=$RUNTIME -DruntimeVersion=$RUNTIME_VERSION --stacktrace --info --no-daemon
+    - mvn install -Dinvoker.streamLogs=true
```
scottkurz: Why does the `boost-maven` part run unconditionally?   Is it because you're thinking ahead to bom/booster use cases?

mattbsox: Yup.

PR #217, opened 2018-11-26
adding support for boosters used in war projects

mattbsox:


boost-gradle/src/main/groovy/io/openliberty/boost/gradle/tasks/BoostPackageTask.groovy
```diff
@@ -92,6 +100,16 @@ public class BoostPackageTask extends AbstractBoostTask {
                             project.bootRepackage.finalizedBy 'boostPackage'
                         }
                     }
+                } else if (isJ2EEProject()) {
```
chyt: At this point you already know it's not a SB project. Does do `else` and remove `isJ2EEProject()` method

boost-gradle/src/main/groovy/io/openliberty/boost/gradle/tasks/BoostPackageTask.groovy
```diff
@@ -104,7 +122,7 @@ public class BoostPackageTask extends AbstractBoostTask {
             //The task will perform this before any other task actions
             doFirst {
 
-                libertyServerPath = "${project.buildDir}/wlp/usr/servers/BoostServer"
+                libertyServerPath = "${project.buildDir}/wlp/usr/servers/${project.liberty.server.name}"
```
scottkurz: We don't have a Maven equivalent for this name, right?  Is this a consequence of the fact that the user configures liberty-gradle directly whereas LMP is wrapped?

boost-common/src/main/java/io/openliberty/boost/common/utils/LibertyServerConfigGenerator.java
```diff
@@ -218,7 +218,11 @@ public void addBootstrapProperties(Properties properties) throws IOException{
     public void addConfigForApp(String artifactId, String version){
     	Element appCfg = doc.createElement(APPLICATION);
     	appCfg.setAttribute("context-root", "/");
-    	appCfg.setAttribute("location", artifactId+"-"+version+ "." + ConfigConstants.WAR_PKG_TYPE);
+        if(version == null) {
+            appCfg.setAttribute("location", artifactId + "." + ConfigConstants.WAR_PKG_TYPE);
+        } else {
+            appCfg.setAttribute("location", artifactId + "-" + version + "." + ConfigConstants.WAR_PKG_TYPE);
```
scottkurz: Maybe a dumb question, but is a 'null' version a thing in Gradle?   It's not in Maven...     Maybe this is just defensive code...

boost-gradle/src/main/groovy/io/openliberty/boost/gradle/tasks/BoostPackageTask.groovy
```diff
@@ -167,6 +191,48 @@ public class BoostPackageTask extends AbstractBoostTask {
...
+            .forComponents(componentIds)
+            .withArtifacts(MavenModule, MavenPomArtifact)
+            .execute()
+
+        for (component in result.resolvedComponents) {
+            logger.info("Adding ${component.id} to dependency list.")
+            component.getArtifacts(MavenPomArtifact).each { dependencyList.add(component.id.toString()) }
+        }
+
+        return boosterParent.mapDependenciesToFeatureList(dependencyList)
```
scottkurz: Maybe slightly better to change signature to return dependencyList, and pass that onto `generateServerXMLJ2EE` ?    The Maven side is doing the same thing...... the core Maven vs. Gradle diff is just building the dependency list.    

Maybe we want to defer building a feature list just yet and wait until we start to generate.

boost-gradle/src/main/groovy/io/openliberty/boost/gradle/tasks/BoostPackageTask.groovy
```diff
@@ -167,6 +191,48 @@ public class BoostPackageTask extends AbstractBoostTask {
...
+            serverConfig.addConfigForFeatures(boosterConfigurators)
+
+            // Write server.xml to Liberty server config directory
+            serverConfig.writeToServer()
+
+        } catch (TransformerException | IOException | ParserConfigurationException e) {
+            throw new GradleException("Unable to generate server configuration for the Liberty server.", e);
+        }
+    }
+
```
scottkurz: This code is pretty key... since it's almost identical w/ Maven side except for exc handling, maybe it should be factored into common as well.

boost-gradle/src/test/groovy/BoostPackageJaxRS20Test.groovy
```diff
@@ -0,0 +1,122 @@
...
+        BuildResult result = GradleRunner.create()
+            .withProjectDir(testProjectDir)
+            .forwardOutput()
+            .withArguments("boostStop", "-i", "-s")
+            .build()
+       
+        assertEquals(SUCCESS, result.task(":boostStop").getOutcome())
+    }
+
+    @Test
```
scottkurz: Are we guaranteed this test runs first or do we need to force somehow?

boost-gradle/src/test/groovy/BoostPackageJaxRS20Test.groovy
```diff
@@ -0,0 +1,122 @@
...
+            }
+        } finally {
+            if (br != null) {
+                br.close()
+            }
+        }
+        
+        assertTrue("The " + JAXRS_20_FEATURE + " feature was not found in the server configuration", found);    
+    }
+
```
scottkurz: Maye worth abstracting into the abstract class?

boost-gradle/src/test/groovy/BoostPackageJaxRS20Test.groovy
```diff
@@ -0,0 +1,122 @@
...
+        try {
+            int statusCode = client.executeMethod(method)
+
+            assertEquals("HTTP GET failed", HttpStatus.SC_OK, statusCode)
+
+            String response = method.getResponseBodyAsString(10000)
+
+            assertTrue("Unexpected response body",
+                    response.contains("Hello World From Your Friends at Liberty Boost EE!"))
+        } finally {
```
scottkurz: That's very friendly :)

boost-gradle/src/main/groovy/io/openliberty/boost/gradle/tasks/BoostPackageTask.groovy
mattbsox: Right, we can just get the properties from liberty-gradle in this case.

boost-common/src/main/java/io/openliberty/boost/common/utils/LibertyServerConfigGenerator.java
mattbsox: The war task can have a null version so when I was testing we were creating an application element with null as part of the app location.

boost-gradle/src/test/groovy/BoostPackageJaxRS20Test.groovy
mattbsox: Should be in same order as they are in this test file.

scottkurz: Is that enforced by Gradle / Groovy specifically?  I know when using plain-old Java in Liberty FAT, the IBM JDK would use declaration order whereas the Oracle JDK would use alpha order.

boost-common/src/main/java/io/openliberty/boost/common/utils/LibertyServerConfigGenerator.java
scottkurz: So this is the version of the war task specifically?  Not the artifact as a whole (as in Maven)?   Maybe this concept doesn't exist in Maven (of the artifact always having a GAV of its own?)  Sorry for my ignorance.

boost-gradle/src/test/groovy/BoostPackageJaxRS20Test.groovy
mattbsox: I'll move the build part into the setup method. Shouldn't matter once it's there.

boost-common/src/main/java/io/openliberty/boost/common/utils/LibertyServerConfigGenerator.java
mattbsox: Right, the project might not have a version assigned to it or the war, which is fine in gradle.

boost-gradle/src/main/groovy/io/openliberty/boost/gradle/tasks/BoostPackageTask.groovy
mattbsox: Ya, there are some other changes we're going to make as part of moving the Booster stuff to common. We can refactor this part over with the other changes.

PR #218, opened 2018-11-28
Add JDBC booster and refactor booster configuration

awisniew90:
1. Refactored Booster logic to common
2. Added booster dependency copy functionality
3. Added JDBC booster with Derby dataSource
4. Added JDBC booster test


boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/liberty/LibertyPackageMojo.java
```diff
@@ -48,11 +42,10 @@
 @Mojo(name = "package", defaultPhase = LifecyclePhase.PACKAGE, requiresDependencyResolution = ResolutionScope.COMPILE_PLUS_RUNTIME, requiresDependencyCollection = ResolutionScope.COMPILE_PLUS_RUNTIME)
 public class LibertyPackageMojo extends AbstractLibertyMojo {
```
scottkurz: Are you using the code formatter?   This looks off a bit.

boost-common/src/main/java/io/openliberty/boost/common/config/JDBCBoosterPackConfigurator.java
```diff
@@ -0,0 +1,97 @@
...
+    	lib.setAttribute("id", "DerbyLib");
+        Element fileLoc = doc.createElement(FILESET);
+        fileLoc.setAttribute("dir", "resources");
+        fileLoc.setAttribute("includes", "derby*.jar");
+        lib.appendChild(fileLoc);
+        serverRoot.appendChild(lib);
+        
+        // Add datasource
+        Element dataSource = doc.createElement(DATASOURCE);    	
+        dataSource.setAttribute("id", "DefaultDerbyDatasource");
```
scottkurz: How did you pick the name ?    I was thinking the special "DefaultDatasource"

boost-maven/boost-boms/boost-ee7-apis-bom/pom.xml
```diff
@@ -24,6 +24,11 @@
 				<artifactId>javax.ws.rs-api</artifactId>
 				<version>2.0</version>
 			</dependency>
+			<dependency>
+			    <groupId>org.apache.derby</groupId>
+			    <artifactId>derby</artifactId>
+			    <version>10.14.2.0</version>
+			</dependency>
 		</dependencies>
```
scottkurz: Why do we want to put this in the BOM?   It's not needed to compile against.    Though it'd be kind of nice to see these internal dependencies in a POM somewhere instead of buried in Java source...not sure we have a good way to do that.

boost-gradle/build.gradle
```diff
@@ -107,6 +107,8 @@ task groovydocJar(type: Jar) {
     from groovydoc
 }
 
+task copyDependency(type: Copy)
```
mattbsox: I think this can be removed.

boost-common/src/main/java/io/openliberty/boost/common/utils/LibertyBoosterUtil.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.List;
+import java.util.Map;
+
+import io.openliberty.boost.common.BoostLoggerI;
+import io.openliberty.boost.common.config.BoosterPackConfigurator;
+import io.openliberty.boost.common.config.JAXRSBoosterPackConfigurator;
+import io.openliberty.boost.common.config.JDBCBoosterPackConfigurator;
+import io.openliberty.boost.common.config.LibertyServerConfigGenerator;
+
+public class LibertyBoosterUtil {
```
mattbsox: I think we should change this class to use static methods like the SpringBootUtil and the other helper classes.

boost-gradle/src/main/groovy/io/openliberty/boost/gradle/utils/GradleProjectUtil.groovy
```diff
@@ -37,4 +40,33 @@ public class GradleProjectUtil {
...
+            }
+        }
+       
+        return springBootDependencies
+    }
+    
+    public static Map<String, String> getBoosterDependencies(Project project) {
+        Map<String, String> boosterDependencies = new HashMap<String, String>()
+
+        project.configurations.compile.resolvedConfiguration.resolvedArtifacts.collect { it.moduleVersion.id }.each { ModuleVersionIdentifier id ->
```
mattbsox: Does this work the same as the ArtifactResolutionQuery we were using before in BoostPackageTask? That was being used to find all of the poms in the project's dependency tree.

boost-maven/boost-maven-plugin/src/it/test-jdbc/src/main/webapp/.gitignore
```diff
@@ -0,0 +1 @@
+/META-INF/
```
mattbsox: Need this file?

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/utils/MavenProjectUtil.java
```diff
@@ -56,5 +58,22 @@ public static String findSpringBootVersion(MavenProject project) {
...
+    public static Map<String, String> getBoosterDependencies(MavenProject project, BoostLogger logger) {
+
+        Map<String, String> boosterDependencies = new HashMap<String, String>();
+        logger.debug("Processing project for dependencies.");
+
+        for (Artifact artifact : project.getArtifacts()) {
+            logger.debug("Found dependency while processing project: " + artifact.getGroupId() + ":"
+                    + artifact.getArtifactId() + ":" + artifact.getVersion());
+            
+            if (artifact.getGroupId().equals("io.openliberty.boosters")) {
```
scottkurz: Should this be a string constant somewhere (not sure where)?

boost-gradle/src/main/groovy/io/openliberty/boost/gradle/utils/GradleProjectUtil.groovy
```diff
@@ -37,4 +40,33 @@ public class GradleProjectUtil {
...
+        
+            if ( id.group.equals("io.openliberty.boosters") ) {
+                
+                boosterDependencies.put(id.name.toString(), id.version.toString())
+            }
```
scottkurz: Could create a new class BoosterArtifact with a `name` and `version`.

boost-common/src/main/java/io/openliberty/boost/common/config/JDBCBoosterPackConfigurator.java
```diff
@@ -0,0 +1,99 @@
...
+    	lib.setAttribute("id", "DerbyLib");
+        Element fileLoc = doc.createElement(FILESET);
+        fileLoc.setAttribute("dir", "resources");
+        fileLoc.setAttribute("includes", "derby*.jar");
+        lib.appendChild(fileLoc);
+        serverRoot.appendChild(lib);
+        
+        // Add datasource
+        Element dataSource = doc.createElement(DATASOURCE);    	
+        dataSource.setAttribute("id", "DefaultDataSource");
```
scottkurz: I'd make a constant out of the strings that aren't "obvious" such as "DefaultDataSource"

boost-common/src/main/java/io/openliberty/boost/common/utils/LibertyBoosterUtil.java
scottkurz: I kind of agree, not just for the sake of consistency but also because it steers us to thinking more carefully about object interactions and lifecycles to give more specific names to things we're creating and holding an instance of... vs. code that just statelessly returns output from input.    If it's really something we want an instance of... we should be able to find a better name than xxxUtil

boost-gradle/src/main/groovy/io/openliberty/boost/gradle/utils/GradleProjectUtil.groovy
```diff
@@ -37,4 +40,33 @@ public class GradleProjectUtil {
...
     }
+    
+    public static List<String> getSpringFrameworkDependencies(Project project) {
+
+        List<String> springBootDependencies = new ArrayList<String>()
+
+        project.configurations.compile.resolvedConfiguration.resolvedArtifacts.collect { it.moduleVersion.id }.each { ModuleVersionIdentifier id ->
+        
+            if ( id.group.equals("org.springframework") ) {
+
```
scottkurz: This isn't a "startsWith"  of "org.springframework"  ???

boost-maven/boost-maven-plugin/src/it/test-jdbc/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+			<artifactId>commons-httpclient</artifactId>
+			<version>3.1</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+            <scope>provided</scope>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
+            <version>3.1.0</version>
+        </dependency>
```
scottkurz: At EE 8 level, we'd probably be using servlet 4.0...can't think of a real concrete problem with using the 3.1 here though.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/liberty/LibertyPackageMojo.java
```diff
@@ -333,6 +299,31 @@ private void installMissingFeatures() throws MojoExecutionException {
...
+    	
+    	List<String> dependenciesToCopy = boosterUtil.getDependenciesToCopy();
+    	
+    		
+		for (String dep :  dependenciesToCopy) {
+			
+			String[] dependencyInfo = dep.split(":");
+			
+			executeMojo(getMavenDependencyPlugin(), goal("copy"), configuration(element(name("outputDirectory"), libertyServerPath + "/resources"),
+	    			element(name("artifactItems"),
```
scottkurz: For Maven and Gradle... is  the server resources directory better .. or the shared resources dir?     Perhaps it doesn't matter since this is targeted to a single server... but we still put the SpringBoot dependencies into shared... IDK.

boost-maven/boost-maven-plugin/src/it/test-jdbc/src/main/webapp/.gitignore
awisniew90: Nope. Got rid of it

boost-gradle/build.gradle
awisniew90: Done

boost-common/src/main/java/io/openliberty/boost/common/utils/LibertyBoosterUtil.java
awisniew90: I think having an instantiation of this class will help reduce the number of times we need to pass the same variables to the methods. It also pushes the resolution of the BoosterConfigurators into common rather than having the gradle and maven plugins doing that. Other than consistency with other helper classes, is there a preference to having static classes?

boost-gradle/src/main/groovy/io/openliberty/boost/gradle/utils/GradleProjectUtil.groovy
awisniew90: I'm fairly certain this is getting all of the dependencies in the poms. I will print them out using both ways and make sure.

mattbsox: Did a bit of testing. Looks good.

boost-common/src/main/java/io/openliberty/boost/common/utils/LibertyBoosterUtil.java
awisniew90: Hey @scottkurz  - Sorry, I commented before seeing your comment. Either that or git is putting out chain out of order.  Which way do you agree with?

boost-gradle/src/main/groovy/io/openliberty/boost/gradle/utils/GradleProjectUtil.groovy
awisniew90: Thanks Matt

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/liberty/LibertyPackageMojo.java
awisniew90: I don't think `server package` will package up what's in the shared resource dir into the uber jar.

boost-maven/boost-maven-plugin/src/it/test-jdbc/pom.xml
awisniew90: I can change and test this in the next item

boost-gradle/src/main/groovy/io/openliberty/boost/gradle/utils/GradleProjectUtil.groovy
awisniew90: I know we addressed this in gitter, but just for consistency here, the spring artifacts we are interested in have "org.springframework" as the full groupID. This is acting as a filter here.

boost-common/src/main/java/io/openliberty/boost/common/config/JDBCBoosterPackConfigurator.java
awisniew90: Will address this in the next item.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/utils/MavenProjectUtil.java
awisniew90: Agree. I will change and test this in the next item

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/liberty/LibertyPackageMojo.java
```diff
@@ -161,6 +161,33 @@ private void generateServerConfigSpringBoot() throws MojoExecutionException {
...
         }
     }
+    
+    /**
+     * Generate config for the Liberty server based on the Maven Spring Boot
+     * project.
+     * 
+     * @throws MojoExecutionException
+     */
+    private void generateServerConfigJ2EE() throws MojoExecutionException {
```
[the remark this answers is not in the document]
awisniew90: Agree

boost-common/src/main/java/io/openliberty/boost/common/utils/LibertyBoosterUtil.java
cherylking: I tend to agree that they should be static. Especially since they are in common and are basically helper methods.

awisniew90: That's fine with me. @mattbsox - I can make this change in the next issue if you are ok with that.  Just to move this issue along for now.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/liberty/LibertyPackageMojo.java
scottkurz: Didn't realize that.

PR #221, opened 2018-12-04
Moving shared package functionality to boost-common

mattbsox:
Created a PackageUtil class in boost-common
Moved shared package functionality to PackageUtil and SpringBootUtil

### DIFF
--- a/boost-common/src/main/java/io/openliberty/boost/common/utils/SpringBootUtil.java
+++ b/boost-common/src/main/java/io/openliberty/boost/common/utils/SpringBootUtil.java
@@ -90,7 +90,7 @@ public class SpringBootUtil {
      *         null otherwise
      * @throws PluginExecutionException
      */
-    public static File copySpringBootUberJar(File artifact, BoostLoggerI logger) throws PluginExecutionException {
+    public static File copySpringBootUberJarFile(File artifact, BoostLoggerI logger) throws PluginExecutionException {
         if (artifact == null || !artifact.exists() || !artifact.isFile()) {
             throw new BoostException("Could not find a project artifact.");
         }
@@ -109,6 +109,44 @@ public class SpringBootUtil {
             }
         }
         return null;
+    }
+
+    /**
+     * Copy the Spring Boot Uber JAR to a .spring extension to preserve it
+     * 
+     * @param springBootUberJar
+     * @param projectArtifact
+     * @param shouldReplaceProjectArchive
+     * @param boostLogger
+     * @throws BoostException
+     */
+
+    public static void copySpringBootUberJar(File springBootUberJar, File projectArtifact, boolean shouldReplaceProjectArchive, BoostLoggerI boostLogger) throws BoostException {
+        try {
+            File springBootUberJarCopy = null;
+            if (springBootUberJar != null) { // Only copy the Uber JAR if it is valid
+                springBootUberJarCopy = SpringBootUtil.copySpringBootUberJarFile(springBootUberJar,
+                        boostLogger);
+            }
+
+            if (springBootUberJarCopy == null) {
+                boostLogger.info("Plugin should replace the project archive: " + shouldReplaceProjectArchive);
+                if (shouldReplaceProjectArchive) {
+                    if (projectArtifact != null) {
+                        File springJar = new File(
+                            getBoostedSpringBootUberJarPath(projectArtifact));
+                        if (net.wasdev.wlp.common.plugins.util.SpringBootUtil.isSpringBootUberJar(springJar)) {
+                            boostLogger.info("Copying back Spring Boot Uber JAR as project artifact.");
+                            FileUtils.copyFile(springJar, projectArtifact);
+                        }
+                    }
+                }
+            } else {
+                boostLogger.info("Copied Spring Boot Uber JAR to " + springBootUberJarCopy.getCanonicalPath());
+            }
+        } catch (Exception e) {
+            throw new BoostException(e.getMessage(), e);
+        }  
     }
 
     /**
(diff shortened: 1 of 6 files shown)


boost-common/src/main/java/io/openliberty/boost/common/utils/SpringBootUtil.java
```diff
@@ -111,6 +111,44 @@ public static File copySpringBootUberJar(File artifact, BoostLoggerI logger) thr
         return null;
     }
 
+    /**
```
cherylking: I think the method comments for this method and the previous one need to contain more details. I'm wondering why we have two methods to copy the Spring Boot Uber JAR. When would I use this one vs the previous one?

boost-gradle/src/main/groovy/io/openliberty/boost/gradle/tasks/BoostPackageTask.groovy
```diff
@@ -173,21 +175,7 @@ public class BoostPackageTask extends AbstractBoostTask {
         return springBootDependencies
     }
 
-    public String getClassifier() {
```
cherylking: Where did this go? Doesn't something use this method?

boost-common/src/main/java/io/openliberty/boost/common/utils/SpringBootUtil.java
chyt: I guess the `copySpringBootUberJarFile` method is really just a helper for `copySpringBootUberJar`. The first method will conditionally perform the actual copy, while the second does the copy back. If you recall, the logic was written this way to allow re-execution of the `package` goal with a Liberty Spring Boot artifact already generated. This means that we need the same "entry point" for both behaviors since we are unsure of the state of the build previous build (if any) at this point in the execution. 

It looks like we don't call `copySpringBootUberJarFile` anywhere outside of this method, so I think it's sufficient to change the visibility to `private` and indicate in the comments that this is a helper for the publicly visible `copySpringBootUberJar`.

boost-common/src/main/java/io/openliberty/boost/common/utils/SpringBootUtil.java
```diff
@@ -111,6 +111,44 @@ public static File copySpringBootUberJar(File artifact, BoostLoggerI logger) thr
...
+                        if (net.wasdev.wlp.common.plugins.util.SpringBootUtil.isSpringBootUberJar(springJar)) {
+                            boostLogger.info("Copying back Spring Boot Uber JAR as project artifact.");
+                            FileUtils.copyFile(springJar, projectArtifact);
+                        }
+                    }
+                }
+            } else {
+                boostLogger.info("Copied Spring Boot Uber JAR to " + springBootUberJarCopy.getCanonicalPath());
+            }
+        } catch (Exception e) {
```
chyt: Can we specifically indicate the dependencies being thrown here? I think it's cleaner than just doing a catch-all.

boost-common/src/main/java/io/openliberty/boost/common/utils/SpringBootUtil.java
```diff
@@ -111,6 +111,44 @@ public static File copySpringBootUberJar(File artifact, BoostLoggerI logger) thr
...
+            throw new BoostException(e.getMessage(), e);
```
chyt: Not very useful here to just propagate the same error message. I would have the same thing here as in line 108:

```
throw new BoostException("Error copying Spring Boot Uber JAR.", e);
```

boost-common/src/main/java/io/openliberty/boost/common/utils/SpringBootUtil.java
```diff
@@ -260,84 +298,95 @@ public static Properties getSpringBootServerProperties(String springBootProjectR
...
+            } else {
+                serverConfig.addHttpEndpoint(makeVariable(SpringBootUtil.SERVER_ADDRESS), makeVariable(SpringBootUtil.SERVER_PORT), null);
+            }
+            
+            // Add properties to bootstrap properties
+            serverConfig.addBootstrapProperties(springBootServerProps);
+            
+            serverConfig.writeToServer();
+
+        } catch (Exception e) {
```
chyt: Same thing here, list all the exception instead of a catch-all.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/liberty/LibertyPackageMojo.java
```diff
@@ -80,64 +74,66 @@ public void execute() throws MojoExecutionException {
...
-            // we install the app now, after server.xml is configured. This is
-            // so that we can specify a custom config-root in server.xml ("/").
-            // If we installed the app prior to server.xml configuration, then
-            // the LMP would write out a webapp stanza into config dropins that
-            // would include a config-root setting set to the app name.
-            installApp(ConfigConstants.NORMAL_PROJ);
-
-            createUberJar(null, attach);
+        } catch (BoostException be) {
+            throw new MojoExecutionException("Encountered an error when packaging the project:.", be);
```
chyt: Remove the colon.

boost-gradle/src/main/groovy/io/openliberty/boost/gradle/tasks/BoostPackageTask.groovy
mattbsox: We weren't using this anywhere so I removed it.

scottkurz:
This ends up overlapping the changes in #218, which we decided on Gitter to merge instead.   We'll close this and note a couple of the review comments in #218 should be followed-up on....

Issue #215, opened 2018-11-20
SpringBoot applications don't honor `-Dserver.port` override from boost:run goal.


scottkurz:
I can do:
  `java -Dserver.port=8081 -jar target/my.jar`

to override the server.port property but if I do:
  ` mvn  boost:run -Dserver.port=8081` 

the override has no effect.

OTOH, if I use a non-Boost SpringBoot app I can do this to override.
` mvn  spring-boot:run -Dserver.port=8081`

scottkurz:
I'd assume the same issue applies when using 'start', 'debug' goals.

Opened this after trying with Maven.  Not sure if this applies for Gradle as well.

PR #223, opened 2018-12-06
Issue 215 Add addtional test for run

uberskigeek:


boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/liberty/AbstractLibertyMojo.java
```diff
@@ -88,5 +100,72 @@ private void createDefaultRuntimeArtifactIfNeeded() {
     public void execute() throws MojoExecutionException {
         createDefaultRuntimeArtifactIfNeeded();
     }
+    /*
+     * Check for environment variable being set for server port. 
+     * if set update bootstrap.properties to add the server port
+     */
+    protected void setServerPort() {
```
scottkurz: I can appreciate why you did it this way.   

This lets you override the port calculated in package during run/start/debug without affecting the port used by the packaged artifact.

However it doesn't allow the same -Dserver.port=8081 override during package.

I'm thinking a better approach would be to incorporate the -D CLI property into the property calculation used in `io.openliberty.boost.common.utils.SpringBootUtil.generateLibertyServerConfig`  and `getSpringBootServerProperties`, and to do this calculation including writing the whole thing out to boostrap.properties every time.  

Also, instead of having a Maven/Gradle-specific second place to adjust properties, let's pass the needed properties in from the build environment in a Maven/Gradle-specific way as needed, and write the properties in one place.

This is a new thing, so let's make sure to discuss with the team and hopefully get at least one review besides myself.

scottkurz:
Going to close this out, unless you object @uberskigeek ...  Seems like this hasn't come up as a use case demanding attention, plus involves some new config like jvm.options, plus the code has diverged a bit.    

And we still have #215 open.

PR #226, opened 2018-12-12
Fix issues with detecting end-of-string in different versions of sed

scottkurz:


pre-commit.template
```diff
@@ -57,7 +57,7 @@ do
         #
         # Note the sed output is a relative path (relative to project location)
         #
```
scottkurz:
```suggestion
        #
        # The idea is this translates
        #  R a/b/c/X.java -> a/b/d/Y.java 
        # into  'a/b/d/Y.java' .   That is, it strips off everything leaving the new relative path.
        # It assumes there are no spaces in the path name.
        #  We had some inconsistencies with Git versions and sed versions, so hopefully this covers all cases.
```

PR #227, opened 2018-12-13
Fix issue where files to format list was getting truncated due to cas…

scottkurz:
…e-sensitive var in hook script


scottkurz:
@uberskigeek - if you look at the diff quick it probably just looks like I have a loveOfCamelCaseOrSomething :)

The problem though was the one var on line 84 that was camelcase while the others were NOT, since this was resulting in a totally separate variable.    But since the other vars in the script were camel case I just fixed this one to be, in making it consistent.

uberskigeek:
That would be a problem!

PR #229, opened 2018-12-13
JDBC booster with a user defined derby dependency

awisniew90:
Have our JDBC booster pick up a user defined Derby dependency


awisniew90:
Note: I ended up rebasing this commit on top of the previous initial JDBC driver support _after_ I formatted using the pre-commit hook... So several of the files have a lot of formatting changes. sorry :-/ . 

Also, in order to detect if a derby dependency was added, I ended up changing how we find our booster dependencies and spring framework dependencies. Now there is just one method in MavenProjectUtil/GradleProjectUtil that just finds and returns a list of all dependencies. This allows us to easily find dependencies like derby, and makes this a little cleaner rather than trying to do several layers of filtering.

scottkurz:
Before we start reviewing this in detail, let me just note it doesnt' seem to be there's any unit test for this, and I don't recall there was for the last one either.    Could we try to at least do one, to build some momentum?  Something simple looking at the generated server config output maybe?

awisniew90:
True, we should probably add some basic unit tests here.  I'll add those in.

awisniew90:
Added unittests for checking JDBC feature version and datasource configuration in server.xml

scottkurz:
@awisniew90 can you please rebase this onto master first (unless there's a specific, temporary reason for keeping the existing commits)?

awisniew90:
@scottkurz - Rebased onto master and got rid of that extra commit that was hanging around.

PR #249, opened 2019-01-16
Add DB2 support to JDBC booster

awisniew90:
This PR adds support for DB2 to the existing JDBC booster. If a db2 maven/gradle dependency is added to a project, the JDBC booster will use that as its datasource. Additionally, 4 new properties were added to support a basic default db2 type 4 connection.

### DIFF
--- a/boost-common/src/main/java/io/openliberty/boost/common/config/JDBCBoosterPackConfigurator.java
+++ b/boost-common/src/main/java/io/openliberty/boost/common/config/JDBCBoosterPackConfigurator.java
@@ -23,16 +23,19 @@ import org.w3c.dom.NodeList;
 
 public class JDBCBoosterPackConfigurator extends BoosterPackConfigurator {
 
-    private static String DERBY_DEFAULT = "org.apache.derby:derby:10.14.2.0";
+	public static String DERBY_DEPENDENCY = "org.apache.derby:derby";
+    public static String DB2_DEPENDENCY = "com.ibm.db2.jcc:db2jcc";
     
-    public static String DEFAULT_DATABASE_NAME = SERVER_OUTPUT_DIR + "/" + DERBY_DB;
+    private static String DERBY_DEFAULT = "org.apache.derby:derby:10.14.2.0";
+    public static String DEFAULT_DERBY_DATABASE_NAME = DERBY_DB;
+    public static String DEFAULT_DB2_DATABASE_NAME = DB2_DB;
 
     private String dependency;
     private String libertyFeature;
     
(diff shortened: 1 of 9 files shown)


boost-common/src/main/java/io/openliberty/boost/common/config/JDBCBoosterPackConfigurator.java
```diff
@@ -89,25 +149,54 @@ public void addServerConfig(Document doc) {
...
 
-        return dependency;
-    }
+        // Add library
+        Element lib = doc.createElement(LIBRARY);
+        lib.setAttribute("id", DB2_LIB);
+        Element fileLoc = doc.createElement(FILESET);
+        fileLoc.setAttribute("dir", RESOURCES);
+        fileLoc.setAttribute("includes", "db2jcc*.jar"); 
+        lib.appendChild(fileLoc);
```
scottkurz: So it's a directory but not everything is included... just `db2jcc*.jar`.    What's the intent here?  Are you supposed to supply a license somewhere (which would match this pattern)?

boost-common/src/main/java/io/openliberty/boost/common/utils/BoostUtil.java
```diff
@@ -75,5 +78,38 @@ public static void extract(File artifact, File projectDirectory, String dependen
     public static String makeVariable(String propertyName) {
         return "${" + propertyName + "}";
     }
+    
+    public static String encrypt(String libertyInstallPath, String password) throws IOException {
+    	
+    	Runtime rt = Runtime.getRuntime();
+    	String[] commands = { libertyInstallPath + "/bin/securityUtility", "encode", password };
+    	Process proc = rt.exec(commands);
```
scottkurz: Is there an API equivalent or do we need a new process?

boost-common/src/main/java/io/openliberty/boost/common/config/JDBCBoosterPackConfigurator.java
awisniew90: @scottkurz - the jar is the only file that is part of the db2 maven dependency. Im not aware of a license.  is one typically needed?

boost-common/src/main/java/io/openliberty/boost/common/utils/BoostUtil.java
awisniew90: There is, but I think it's only available if you are running in a Liberty server. I'm not sure how to access it from Boost code.

Issue #252, opened 2019-01-22
Handle passwords that are already encrypted


mattbsox:
We need to determine if we have a password that is already encrypted in a properties file or from the command line before it gets passed to the securityUltility for encryption.

scottkurz:
Fixed by #255

PR #255, opened 2019-01-23
Adding check for already encoded DB password

mattbsox:
- Checking error message for an InvalidPasswordEncodingException
- Refactored LibertyServerConfigGenerator, BoostUtil, SpringBootUtil, and LibertyBoosterUtil to use BoostLoggers
- Added test to check properties written to bootstrap.properties


mattbsox:
Fixes #252

PR #256, opened 2019-01-24
Updating boost-gradle to run on Gradle 5.1.1

### LINKED ISSUE #254: Spring 1.5 tests failing when run with Gradle 5.x
The Spring 1.5 Gradle plugin will not be supporting Gradle 5.x. We'll need to add a 4.x wrapper to run the tests with.

We should also document this somewhere in the repo and add a warning that gets displayed with in the Gradle plugin.

mattbsox:
Changed Spring Boot 1.5 tests to use Gradle 4.10 wrapper rather than the project wrapper
Updated Spring Boot 1.5 Gradle plugin verison
Updated the boost-gradle wrapper version from 4.8 to 5.1.1

Fixes #254


mattbsox:
I'll add something to the boost-gradle readme about that.

scottkurz:
Dumb question, but is this only relevant when building/developing boost ?  (As opposed to using boost)

mattbsox:
Different Gradle versions can behave differently which can cause some plugins to break when they're used. Just making sure we're good to run on 5.x. The reason for all the commits is that Travis keeps killing off a daemon while the tests are running so I'm trying to figure that out right now.

scottkurz:
Weird.. my question comes from the fact I must have somehow been looking at only one of your 8 commits rather than the full set.   When I saw what seemed to be your change I asked if it was only relevant when building the boost plugin..but nevermind now.  Thx.

PR #257, opened 2019-01-25
Use reflection to find boosters

chyt:
With this PR, each booster must:
- exist in the `io.openliberty.boost.common.boosters` package
- extend `AbstractBoosterConfig`
- have a constructor with signature `Map<String, String> dependencies, BoostLoggerI logger`
- be annotated with `@BoosterCoordinates(<group-id:artifact-id>)`

This PR also fixes a bug with system properties not being unset in the boost-common JUnit tests. For tests that need to manipulate system properties, add the following rule to ensure that the original system properties are restored after the end of the test execution:

```
@Rule
public final RestoreSystemProperties restoreSystemProperties = new RestoreSystemProperties();
```


boost-common/src/main/java/io/openliberty/boost/common/config/BoosterConfigurator.java
```diff
@@ -0,0 +1,107 @@
...
+            List<AbstractBoosterConfig> boosterPackConfigurators, String warName, BoostLoggerI logger) throws Exception {
+
+        LibertyServerConfigGenerator serverConfig = new LibertyServerConfigGenerator(libertyServerPath, logger);
+
+        // Loop through configuration objects and get features and XML config
+        // (if any)
+        for (AbstractBoosterConfig configurator : boosterPackConfigurators) {
+            serverConfig.addFeature(configurator.getFeature());
+            serverConfig.addBoosterConfig(configurator);
+        }
```
ajm01: One of the items I was working on for the mpConfig support came from a comment about collapsing these two separate calls into one action - as they both are adding to the server.xml. addFeature is a local method that adds the feature string retrieved by the configurator to the server.xml doc object as a feature stanza in the featureManager block. addBoosterConfig is a local method that takes a configurator object and in turn calls that configurator object with a reference to the same doc object to be updated with that configurator's config stanzas. Should both the configurator's feature name and its config stanzas (if there are any) be written out at the same time in a single call?

ajm01: One of the items I was working on for the mpConfig support came from a comment about collapsing these two separate calls into one action - as they both are adding to the server.xml. addFeature is a local method that adds the feature string retrieved by the configurator to the server.xml doc object as a feature stanza in the featureManager block. addBoosterConfig is a local method that takes a configurator object and in turn calls that configurator object with a reference to the same doc object to be updated with that configurator's config stanzas. Should both the configurator's feature name and its config stanzas (if there are any) be written out at the same time in a single call?

chyt: Sorry I'm a bit confused. I don't believed I changed the implementation of this method (it's showing up as additions because git didn't pick up that this file had been renamed). I'm not sure if I appreciate the difference in combining the two calls vs keeping them separate...but from my understanding I don't see a problem the way it is now (one call to add the feature, and another to add additional config). Although maybe `addBoosterConfig` should be taking in `configurator.getServerProperties()` instead of the entire `configurator` object.

Anyways not sure how much this relates to my PR, did you have any changes that you wanted me to make here?

ajm01: you're correct - it doesn't have anything to do with the changes you made - i was thinking proactively and meant this more of a point of discussion. I should have made it is a comment vs a review to get your thoughts. Your response makes sense, I can mark it resolved.

chyt: Sure, no worries :)

boost-common/src/test/java/io/openliberty/boost/common/config/BoosterConfiguratorTest.java
```diff
@@ -8,78 +8,82 @@
...
-public class LibertyBoosterUtilTest {
+import io.openliberty.boost.common.config.BoosterConfigurator;
+import io.openliberty.boost.common.utils.BoosterUtil;
+import io.openliberty.boost.common.utils.CommonLogger;
+
+public class BoosterConfiguratorTest {
+    
+    @Rule
+    public final RestoreSystemProperties restoreSystemProperties = new RestoreSystemProperties();
+
```
scottkurz: What's this  RestoreSystemProperties  for?   Can you maybe add a comment explaining?

scottkurz: Possibly could be post-merge...(next PR).

chyt: This is documented in the PR description, but sure will add a comment here as well.

boost-common/src/main/java/io/openliberty/boost/common/boosters/AbstractBoosterConfig.java
```diff
@@ -9,19 +9,43 @@
...
+public abstract class AbstractBoosterConfig {
+    
+    @Retention(RetentionPolicy.RUNTIME)
+    @Target(ElementType.TYPE)
+    @Inherited
+    public @interface BoosterCoordinates {
+       String value();
+    }
+    
+    public static String getCoordindates(Class<?> klass) throws BoostException {
```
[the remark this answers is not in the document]
chyt: Yup, will fix this in the next PR.

PR #259, opened 2019-01-31
Base jdbc feature on compiler target

awisniew90:


boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/utils/MavenProjectUtil.java
```diff
@@ -51,4 +55,42 @@ public static String findSpringBootVersion(MavenProject project) {
...
+                    Xpp3Dom target = config.getChild("target");
+
+                    if (target != null) {
+                        return target.getValue();
+                    }
+                }
+            }
+        }
+
+        return "1.6";
```
chyt: How do we know that this is the default value?

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/utils/MavenProjectUtil.java
```diff
@@ -51,4 +55,42 @@ public static String findSpringBootVersion(MavenProject project) {
...
+            String propertyValue = property.getValue().toString();
+
+            if (propertyKey.equals("maven.compiler.target") || propertyKey.equals("maven.compiler.release")) {
+                return propertyValue;
+            }
+        }
+
+        List<Plugin> plugins = project.getBuildPlugins();
+
+        for (Plugin plugin : plugins) {
```
chyt: Refactor the plugin configuration retrieval to use [`MavenProjectUtil`](https://github.com/WASdev/ci.maven/blob/master/liberty-maven-plugin/src/main/java/net/wasdev/wlp/maven/plugins/utils/MavenProjectUtil.java) instead.

chyt:
I'm a bit confused about how this works with our booster boms? From my understanding of the implementation, looks like you are basically ignoring the JDBC version resolved from the pom and instead replacing it with the JDK version that you detect? I know we still haven't finalized how we're doing to do booster versioning, but in this case we could possibly just have a single JDBC booster across all specs, since there are no compile dependencies and the "real" version is resolved internally. Anyways, just wanted to make sure we didn't forget about this aspect of the implementation. Otherwise looks good aside from my two comments.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/utils/MavenProjectUtil.java
awisniew90: Ah good catch... the default is 1.6 for maven-compiler-plugin version  3.8.0. (latest version).  previous versions should be 1.5.  I'll have to check to see what maven version they are using.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/utils/MavenProjectUtil.java
awisniew90: Nice - will do.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/utils/MavenProjectUtil.java
awisniew90: Thinking about this more, I don't think it's worth going below 1.6. I believe Liberty requires a minimum jre of 1.6, so other things might break if the app is run on 1.5 or lower.

cherylking: Actually, I think the minimum  is Java 7 now. For Open Liberty it may even be Java 8.

awisniew90: I changed this method to just return the compiler target if it is explicitly configured. Otherwise it will return null.  The JDBC booster will then determine what feature to use. That's also changed to use jdbc-4.1 (SE 7 spec) as a default or for any compiler target value less than 1.7. It's unlikely that someone would be using a target of less than this, and even still, since Liberty wont run (or at least isnt supported) at less than 1.7, there should be little risk to having jdbc-4.1 as a minimum.

awisniew90:
The JDBC dependency version is remaining the same internally. Instead, the compiler target version is now being set as an internal boost property. The JDBC booster will reference this property to determine the jdbc feature to set rather than the dependency version.  Advantages to this are that the boost property can be referenced by other boosters (if needed) and the dependency version is consistent with other boosters.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/liberty/LibertyPackageMojo.java
```diff
@@ -119,6 +119,10 @@ public void execute() throws MojoExecutionException {
 
             // Get booster dependencies from project
             Map<String, String> dependencies = MavenProjectUtil.getAllDependencies(project, BoostLogger.getInstance());
+            
+            // Determine the Java compiler target version and set this internally 
+            String javaCompilerTargetVersion = MavenProjectUtil.getJavaCompilerTargetVersion(project);
+            System.setProperty("boost.internal.compiler.target", javaCompilerTargetVersion);
```
chyt: Also use the `BoostProperties` constant here.

awisniew90: Done

PR #263, opened 2019-02-02
172 no package for war packaging

scottkurz:


mattbsox:
Do we want to add this functionality in boost-gradle as well?

scottkurz:
Yeah let's fix Gradle or at least open another issue.

And we should open two issues:

1. package the JAR for JAR packaging type (for EE/MP)
1. provide an optional way to package the JAR for WAR packaging type.

scottkurz:
Thx @mattbsox ..cherry picked that one commit.

scottkurz:
@mattbsox  - Not sure I need to know or should be asking about code that's now removed.. BUT... I'm curious why the JAR was built in the root dir rather than build/libs, since it won't get cleaned..... was that just maybe something done w/o much thought or was there a reason?

mattbsox:
Looks like we're getting the archive name(which gets used as the path later) from the war.archiveName property. Unless we have the packaging properties set in the boost plugin configuration we're packaging the runnable jar in the projectDir rather than build/libs. We can change where we grab the path from war.archiveName to war.archivePath and get the jar to go where ever the war gets put or just set it to go into build/libs everytime.

scottkurz:
>We can change where we grab the path from war.archiveName to war.archivePath and get the jar to go where ever the war gets put or just set it to go into build/libs everytime.

Thanks Matt.. sounds like maybe we hadn't put too much thought into that before... which isn't unjustified as here we are removing this code (for now).. but could do whatever later.  Appreciate your comment.

PR #267, opened 2019-02-07
add mp booster and related unit tests

ajm01:
added unit test to test microprofile and related boosters (cdi and jsonp)

### DIFF
--- a/boost-common/src/test/java/io/openliberty/boost/common/boosters/MPOpenTracingBoosterTest.java
+++ b/boost-common/src/test/java/io/openliberty/boost/common/boosters/MPOpenTracingBoosterTest.java
@@ -1,0 +1,69 @@
+package io.openliberty.boost.common.boosters;
+
+import static io.openliberty.boost.common.config.ConfigConstants.*;
+import static io.openliberty.boost.common.utils.DOMUtils.getDirectChildrenByTag;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.transform.TransformerException;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.contrib.java.lang.system.RestoreSystemProperties;
+import org.junit.rules.TemporaryFolder;
+import org.w3c.dom.Element;
+
+import io.openliberty.boost.common.BoostException;
+import io.openliberty.boost.common.BoostLoggerI;
+import io.openliberty.boost.common.config.BoostProperties;
+import io.openliberty.boost.common.config.BoosterConfigurator;
+import io.openliberty.boost.common.config.LibertyServerConfigGenerator;
+import io.openliberty.boost.common.utils.BoostUtil;
+import io.openliberty.boost.common.utils.BoosterUtil;
+import io.openliberty.boost.common.utils.CommonLogger;
+import io.openliberty.boost.common.utils.ConfigFileUtils;
+
+public class MPOpenTracingBoosterTest {
+
+    @Rule
+    public TemporaryFolder outputDir = new TemporaryFolder();
+
+    @Rule
+    public final RestoreSystemProperties restoreSystemProperties = new RestoreSystemProperties();
+
+    BoostLoggerI logger = CommonLogger.getInstance();
+
+    /**
+     * Test that the mpOpenTracing-1.1 feature is added to server.xml when the
+     * MPOpenTracing booster version is set to 0.2-SNAPSHOT
+     * 
+     */
+    @Test
+    public void testMPOpenTracingBoosterFeature_MP20() throws Exception {
+
+        LibertyServerConfigGenerator serverConfig = new LibertyServerConfigGenerator(
+                outputDir.getRoot().getAbsolutePath(), logger);
+
+        List<AbstractBoosterConfig> boosters = BoosterConfigurator.getBoosterConfigs(
+                BoosterUtil.createDependenciesWithBoosterAndVersion(MPOpenTracingBoosterConfig.class, "0.2-SNAPSHOT"),
+                logger);
+
+        serverConfig.addFeature(boosters.get(0).getLibertyFeature());
+        serverConfig.writeToServer();
+
+        String serverXML = outputDir.getRoot().getAbsolutePath() + "/server.xml";
+        boolean featureFound = ConfigFileUtils.findStringInServerXml(serverXML,
+                "<feature>" + MPOPENTRACING_11 + "</feature>");
+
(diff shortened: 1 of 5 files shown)


boost-common/src/main/java/io/openliberty/boost/common/boosters/CDIBoosterConfig.java
```diff
@@ -28,11 +29,8 @@
 
     public CDIBoosterConfig(Map<String, String> dependencies, BoostLoggerI logger) throws BoostException {
         String version = dependencies.get(getCoordindates(this.getClass()));
-        
-        // if it is the 1.0 version = EE7 feature level
-        if (version.equals(MP_20_VERSION)) {
-            libertyFeature = CDI_20;
-        }
+        Properties boostConfigProperties = BoostProperties.getConfiguredBoostProperties(logger);
```
chyt: What is the idea in getting the `boostConfigProperties` in every booster? I don't see any of them being used right now.

boost-maven/boost-boms/boost-ee8-apis-bom/pom.xml
```diff
@@ -24,6 +24,16 @@
 				<artifactId>javax.ws.rs-api</artifactId>
 				<version>2.1</version>
 			</dependency>
+			<dependency>
```
chyt: Fix formatting here.

boost-common/src/main/java/io/openliberty/boost/common/boosters/CDIBoosterConfig.java
ajm01: ok, i re-worked all of this onto the latest master, i think all will be resolved, but you can re-look at it if need be.

boost-common/src/test/java/io/openliberty/boost/common/utils/BoosterUtil.java
```diff
@@ -5,18 +5,48 @@
...
+import io.openliberty.boost.common.boosters.JSONPBoosterConfig;
+import io.openliberty.boost.common.boosters.MPConfigBoosterConfig;
+import io.openliberty.boost.common.boosters.MPHealthBoosterConfig;
+import io.openliberty.boost.common.boosters.MPOpenTracingBoosterConfig;
+import io.openliberty.boost.common.boosters.MPRestClientBoosterConfig;
 
 public class BoosterUtil {
-    
-    private static Map<String, String> createDependenciesWithBoosterAndVersion(Class<?> booster, String version) throws BoostException {
+
```
scottkurz: If BoosterUtil is only used in test, should we move it into src/test/java to make it clearer that these are not APIs used by the plugin during a build?

It seems like the naming is off too.. if we're going to have 0.1 and 0.2 for EE 7 and EE 8 for, say, CDI... then `getCDIDependency` is the wrong name...   Maybe just get rid of this name and call `createDependenciesWithBoosterAndVersion` with the arguments each time ?   Or I guess if JDBC is a special case you could leave it but don't follow this pattern.

boost-maven/boost-boms/boost-ee8-apis-bom/pom.xml
```diff
@@ -36,7 +36,17 @@
...
+	            <artifactId>cdi-api</artifactId>
+				<version>2.0</version>
+			</dependency>
+			<dependency>
+    			<groupId>javax.json</groupId>
+    			<artifactId>javax.json-api</artifactId>
+    			<version>1.1</version>
+			</dependency>
+		</dependencies>            
     </dependencyManagement>
```
scottkurz: Adding CDI, JSONP to the EE 8 BOMs probably made sense back when this PR started, but seems kind of like an odd fit given what we've done in the meantime since now these are defined in the MP2.x BOM.  

So the value add of this now would be in non-MP CDI apps with EE 8 as the "platform" (BOM).    That's important, but it probably is worth adding some IT.   Possibly also worth doing EE 7 at the same time since there's a good chance we'd do a shortcut and just use the same test (not to discourage someone from differentiating though if that can be done easily enough).

boost-common/src/test/java/io/openliberty/boost/common/utils/BoosterUtil.java
scottkurz: Why did I think this wasn't already in 'test'?  No idea..sorry

boost-maven/boost-boms/boost-ee8-apis-bom/pom.xml
ajm01: i am going to remove these from the ee8 bom and do this work under a new PR

boost-common/src/test/java/io/openliberty/boost/common/utils/BoosterUtil.java
ajm01: i don't think there is anything special about jdbc, but just to be sure i am going to leave that one as is and make the suggested changes for the ones i introduced via this pr.

boost-common/src/test/java/io/openliberty/boost/common/boosters/MPOpenTracingBoosterTest.java
```diff
@@ -0,0 +1,79 @@
...
+    public TemporaryFolder outputDir = new TemporaryFolder();
+
+    @Rule
+    public final RestoreSystemProperties restoreSystemProperties = new RestoreSystemProperties();
+
+    BoostLoggerI logger = CommonLogger.getInstance();
+
+    /**
+     * Test that the mpHealth-1.0 feature is added as the default when the
+     * Microprofile target is set to 2.0 (1.6)
```
scottkurz: Seems like this comment is wrongly copied/pasted.

boost-common/src/test/java/io/openliberty/boost/common/boosters/MPOpenTracingBoosterTest.java
```diff
@@ -0,0 +1,79 @@
...
+     * @throws TransformerException
+     * @throws IOException
+     * @throws BoostException
+     * @throws SecurityException
+     * @throws NoSuchMethodException
+     * @throws InvocationTargetException
+     * @throws IllegalArgumentException
+     * @throws IllegalAccessException
+     * @throws InstantiationException
+     */
```
scottkurz: If we're not Javadoc-documenting any of these exceptions, maybe would be better to remove all these and make it shorter?

boost-common/src/test/java/io/openliberty/boost/common/boosters/MPOpenTracingBoosterTest.java
scottkurz: What's the `2.0 (1.6)` referring to?  Something from the pre-Boost legacy?

boost-maven/boost-boms/booster-ee8-bom/pom.xml
```diff
@@ -59,6 +59,16 @@
 				<version>0.2-SNAPSHOT</version>
 				<scope>provided</scope>
 			</dependency>
+		    <dependency>
```
scottkurz: Should this all be pulled out (since it's not being tested here in this PR)?

PR #269, opened 2019-02-11
Adding properties for encryption types

### LINKED ISSUE #253: Add option for different encoding types when encoding passwords
The securityUtility has the ability to encode passwords with `--encoding=<xor, aes, or hash>`. We should have some way to configure this when the user wants to encode a password.

mattbsox:
Adding command line properties to specify encryption method used by the securityUtility.
- `boost.pkg.prop.encryption` can be `xor`, `hash`, or `aes`.
- `boost.pkg.prop.encryption.key` is used to specify the key for `aes` encryption.

Fixes #253


boost-common/src/main/java/io/openliberty/boost/common/config/BoostProperties.java
```diff
@@ -25,6 +25,8 @@
     public static final String DATASOURCE_PORT_NUMBER = "boost.db.portNumber";
     public static final String DATASOURCE_USER = "boost.db.user";
     public static final String DATASOURCE_PASSWORD = "boost.db.password";
+    public static final String PASSWORD_ENCRYPTION_METHOD = "boost.pkg.prop.encryption";
```
awisniew90: Minor comment... we might want to change the name of the external property to something like "boost.password.encryption" or something like that. Some indicator that this is for password encryption?

mattbsox: Ya, think it should still be related to the db password though? Maybe "boost.db.password.encryption"?

awisniew90: Hmm good question. Depends if we want to have each type of password have a different encryption or if we want all passwords configured in boost to use the same type.  I'd think we want a generic "encrypt all passwords with this encoding" property, so maybe leave off the "db" part of the name.  But @scottkurz or @cherylking might have a better idea.

scottkurz:
For the AES option, we should consider how to follow the guidelines [here](https://www.ibm.com/support/knowledgecenter/SSRTLW_8.5.5/com.ibm.websphere.wlp.nd.multiplatform.doc/ae/cwlp_pwd_encrypt.html).   

E.g. see: 

> The encryption key used for decrypting can be overridden from the default by setting the wlp.password.encryption.key property. This property should not be set in the server.xml file that stores the password, but in a separate configuration file that is included by the server.xml file. This separate configuration file should contain only a single property declaration, and should be stored outside the normal configuration directory for the server. This ensures that the file containing the key is not included when you are running the server dump or package command. The encryption key property can also be specified as a bootstrap property. If you choose this option, you should put the encryption key in a separate properties file that is included in the server bootstrap.properties file.

For the 'hash' option, how do we recover the original value?

mattbsox:
Good point. I'll look into splitting up that config a little more.

For the hash option, we can't get the original value back. It shouldn't really be an issue since you wouldn't use it for anything aside from generating the hashes that you'll use for access.

mattbsox:
Okay, I was wrong about this. Looks like we should be using aes and not including hash at all for DB related encryption.

scottkurz:
@mattbsox  - Was wondering if we should start over on this one.  It's got merge conflicts from new code that's been merged, and our original design has changed.   I don't think the review history itself is adding too much value either.   

What do you think about closing this and opening a new PR against https://github.com/OpenLiberty/boost/issues/253 ?

mattbsox:
Sounds good to me.

PR #270, opened 2019-02-15
Add JPA booster

### LINKED ISSUE #136: Create a JPA booster for Java EE 7
This booster should install the correct feature and configure it in the Liberty server.xml.

### LINKED ISSUE #137: Create a JPA booster for Java EE 8
This booster should install the correct feature and configure it in the Liberty server.xml.

chyt:
- JPA 2.1 test also tests defining just JPA booster but getting both JPA and JDBC features
- JPA 2.2 test also tests defining JPA and JDBC boosters but only getting one feature block each of JPA and JDBC features
- Right now JPA boosters are defining both API and booster BOMs, will look into combining them in a separate PR

Closes #136 and closes #137.


chyt:
I'll keep the booster/API boms separate for now while we finalize the decision on whether boosters should list dependencies for other boosters.

boost-maven/boost-boms/pom.xml
```diff
@@ -14,7 +14,6 @@
         <version>0.1.3-SNAPSHOT</version>
     </parent>
 	
-	<groupId>io.openliberty.boost</groupId>
```
awisniew90: Just curious - is this removed because the group id is inherited from the parent?

boost-maven/boost-maven-plugin/src/it/test-jpa-2.2/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+	</dependencyManagement>
+
+	<dependencies>
+		<dependency>
+			<groupId>io.openliberty.boosters</groupId>
+			<artifactId>jpa</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>io.openliberty.boosters</groupId>
+			<artifactId>jdbc</artifactId>
```
awisniew90: Do we need JDBC defined here?

boost-gradle/src/test/resources/test-jpa/src/main/resources/META-INF/persistence.xml
```diff
@@ -0,0 +1,28 @@
...
+    <non-jta-data-source>java:comp/env/jdbc/samplejpadatasourcenonjta</non-jta-data-source>
+
+    <class>application.servlet.Thing</class>
+    <exclude-unlisted-classes>true</exclude-unlisted-classes>
+    <properties>
+      <!-- These properties are creating the database on the fly. We are using them to avoid users having
+          to create a database to run the sample. 
+          See also the create=true line in the datasource meta data. -->
+      <property name="openjpa.jdbc.SynchronizeMappings" value="buildSchema(ForeignKeys=true)" />
+      <property name="openjpa.jdbc.DBDictionary" value="derby" />
```
awisniew90: My JPA knowledge is pretty slim, but how does this relate to the JDBC booster also creating a DB?  Or is that DB not used?

boost-maven/boost-boms/pom.xml
chyt: Yep, and the version is as well.

boost-maven/boost-maven-plugin/src/it/test-jpa-2.2/pom.xml
chyt: As mentioned in the PR description, I'm trying to test that the boosters are configured the same when jdbc is added implicitly or explicitly.

boost-gradle/src/test/resources/test-jpa/src/main/resources/META-INF/persistence.xml
chyt: That's a good point. I took this `persistence.xml` from one of our sample projects on wasdev and I didn't notice this. Based on my experimentation it looks like the application is using the JDBC driver (it fails without the driver config) but possibly not the default database we are configuring. I'll look into this more.

chyt: This is fixed in https://github.com/OpenLiberty/boost/pull/270/commits/9d03eaec65bfd917d8a78ecc48fdcc0051a2c43f.

PR #280, opened 2019-03-07
Dev/Release project support for maven

awisniew90:
The two main changes with this are:
1. Allowing multiple war-type dependencies to be specified
2. Importing boosters from war-type dependencies


boost-common/src/main/java/io/openliberty/boost/common/config/BoosterConfigurator.java
```diff
@@ -77,11 +77,13 @@ public static void generateLibertyServerConfig(String libertyServerPath,
...
-        if (warName != null) {
-            serverConfig.addApplication(warName);
+        if (!warNames.isEmpty()) {
+        	for (String warName : warNames) {
+        		serverConfig.addApplication(warName);
+        	}
         } else {
             throw new Exception(
-                    "Unsupported packaging type - Liberty Boost currently supports WAR packaging type only.");
+                    "No war files were found. The project must have a war pacakging type or specify war dependencies.");
```
chyt: Actually just noticed a typo here: `pacakging`

awisniew90: 👍

PR #281, opened 2019-03-07
Property Encoding for DB passwords

mattbsox:
Adding ability to encode DB passwords with AES.

- Will use AES to encode the `boost.db.password` property when writing the property to bootstrap.properties.
- Will not encode passwords with an existing `{aes}` flag in the passed in password, but will write it straight to boostrap.properties instead.
- An encoding key can be provided through the `boost.db.password.key` property.


boost-common/src/main/java/io/openliberty/boost/common/utils/BoostUtil.java
```diff
@@ -88,38 +90,46 @@ private static String getSecurityUtilCmd(String libertyInstallPath) {
         }
     }
 
-    public static String encrypt(String libertyInstallPath, String password, BoostLoggerI logger) throws IOException {
+    public static String encrypt(String libertyInstallPath, String password, String encryptionKey, BoostLoggerI logger) throws IOException {
+        //Won't encode the password if it contains the aes flag
+        if (!password.contains("{aes}")) {
```
cherylking: What if the user provided an encrypted password that used something other than AES? Do we really want to encrypt it again using AES?

boost-common/src/main/java/io/openliberty/boost/common/config/BoostProperties.java
```diff
@@ -29,6 +29,8 @@
     public static final String DATASOURCE_PASSWORD = "boost.db.password";
     public static final String DATASOURCE_CREATE_DATABASE = "boost.db.createDatabase";
 
+    public static final String PASSWORD_ENCRYPTION_KEY = "boost.db.password.key";
+
```
cherylking: This looks like it is being used to encrypt all passwords (presuming there will be properties for other boosters that require passwords). At line 228 in LibertyServerConfigGenerator, this constant is used to encrypt all password properties. So I question the value of "boost.db.password.key"...should it just be "boost.password.encryption.key"?

awisniew90: Also - we are keying off of the "boost.db." prefix to auto generate variables on the datasource property elements in server.xml. If we do decide to have this password key specific to the datasource password, we would need to change the naming scheme, or make sure we don't generate a variable called "password.key".

boost-common/src/main/java/io/openliberty/boost/common/utils/BoostUtil.java
mattbsox: Added a check for the 3 encryption types we support.

boost-common/src/main/java/io/openliberty/boost/common/config/BoostProperties.java
mattbsox: I added an encryption type value along with the properties we want to encrypt in BoostProperties.

I also changed the name of the aes key property to `boost.aes.key`. It will be used for all of the aes encoding we do.

PR #282, opened 2019-03-08
Dev/Release project support for Gradle

mattbsox:
Added the `boostApp` configuration which can be used to specify files or Maven artifacts to be installed.
Will pull in transitive dependencies from Maven artifacts.
Currently only supports wars.


boost-gradle/src/main/groovy/io/openliberty/boost/gradle/tasks/BoostPackageTask.groovy
```diff
@@ -195,13 +196,15 @@ public class BoostPackageTask extends AbstractBoostTask {
...
-        if (project.war != null) {
+        if (project.plugins.hasPlugin('war')) {
 
             if (project.war.version == null) {
                 warName = project.war.baseName
             } else {
                 warName = project.war.baseName + "-" + project.war.version
             }
+        } else {
+            warName = getWarNameFromBoostApps()
```
chyt: Should we be able to install multiple wars if we find them? Also install war both from the project and dependencies?

boost-gradle/src/main/groovy/io/openliberty/boost/gradle/utils/GradleProjectUtil.groovy
```diff
@@ -45,13 +47,33 @@ public class GradleProjectUtil {
...
+            logger.debug("The compileClasspath configuration was not found.")
+        }
+
+        //Will always have the boostApp configuration since we create it in apply()
+        dependencies.putAll(getAllDependenciesFromConfiguration(project.configurations.boostApp, logger))
+
+        return dependencies
+    }
+
+    public static Map<String, String> getAllDependenciesFromConfiguration(Configuration configuration, BoostLogger logger) {
```
chyt: I think this method should be private since it is only called from `getAllDependencies`.

boost-gradle/src/test/resources/devReleaseApp/dev/build.gradle
```diff
@@ -0,0 +1,50 @@
...
+    imports {
+          mavenBom "io.openliberty.boosters:ee8-bom:${boosterVersion}"
+    }
+}
+
+dependencies {
+	compile "io.openliberty.boosters:jaxrs"
+	compile "io.openliberty.boosters:jdbc"
+	compile("org.apache.httpcomponents:httpclient:4.5.6")
+    compile("org.apache.httpcomponents:httpcore:4.4.10")
```
chyt: Fix indent

boost-gradle/src/test/resources/devReleaseApp/dev/src/main/webapp/.gitignore
```diff
@@ -0,0 +1 @@
+/META-INF/
```
chyt: Can delete this file.

boost-gradle/src/main/groovy/io/openliberty/boost/gradle/utils/GradleProjectUtil.groovy
```diff
@@ -45,13 +47,33 @@ public class GradleProjectUtil {
...
+        Map<String, String> dependencies = new HashMap<String, String>()
+        configuration.resolvedConfiguration.resolvedArtifacts.collect { it.moduleVersion.id }.each { ModuleVersionIdentifier id ->
         	logger.debug("Found dependency while processing project: " + id.group.toString() + ":"
                     + id.name.toString() + ":" + id.version.toString())
                     
             dependencies.put(id.group.toString() + ":" + id.name.toString(), id.version.toString())
```
awisniew90: I think we will only want to add the "booster" dependencies of the boostApp. Any non-booster dependency of the build project (like a db2 or mysql driver) should't be pulled in since it will conflict with what the dev project might be using. This is just my opinion and what I added into the maven build/release support, but I think additional runtime dependencies like that should be determined by the release project.

boost-gradle/src/main/groovy/io/openliberty/boost/gradle/tasks/BoostPackageTask.groovy
mattbsox: Changed this to use a list of war names.

boost-gradle/src/main/groovy/io/openliberty/boost/gradle/utils/GradleProjectUtil.groovy
mattbsox: Updated this to only return the boosters.

PR #289, opened 2019-03-29
Add http endpoint properties

awisniew90:
1. Create properties for boost.hostname, boost.http.port, and boost.https.port.
2. Configure Liberty defaultHttpEndpoint with bootstrap variables
3. Set variables in bootstrap.properties for all three values (either user configured or default values) This allows these values to always be overridable at runtime, even if they were not configured at package time.

4. One possible limitation is in how these properties are propagated to integration tests. Currently, you can set a system property under the failsafe plugin configuration that points to boost.http.port (for example), but this value will only be set if a -Dboost.http.port is specified on "mvn package". Setting boost.http.port as a pom property doesnt seem to work, as the maven pom properties are not set as system properites, and boost will not be aware of the property to configure the server with.  This might just be a limitation with how we allow users to configure boost properties at the moment.

### DIFF
--- a/boost-common/src/main/java/io/openliberty/boost/common/config/BoosterConfigurator.java
+++ b/boost-common/src/main/java/io/openliberty/boost/common/config/BoosterConfigurator.java
@@ -15,6 +15,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Properties;
 import java.util.Set;
 
 import org.reflections.Reflections;
@@ -22,6 +23,7 @@ import org.reflections.Reflections;
 import io.openliberty.boost.common.BoostException;
 import io.openliberty.boost.common.BoostLoggerI;
 import io.openliberty.boost.common.boosters.AbstractBoosterConfig;
+import io.openliberty.boost.common.utils.BoostUtil;
 
 public class BoosterConfigurator {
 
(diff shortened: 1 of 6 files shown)


boost-common/src/main/java/io/openliberty/boost/common/config/BoosterConfigurator.java
```diff
@@ -78,7 +80,26 @@ public static void generateLibertyServerConfig(String libertyServerPath,
...
+        
+        String http_port = (String) boostConfigProperties.getOrDefault(BoostProperties.ENDPOINT_HTTP_PORT, "9080");
+        endpointProperties.put(BoostProperties.ENDPOINT_HTTP_PORT, http_port);
+        
+        String https_port = (String) boostConfigProperties.getOrDefault(BoostProperties.ENDPOINT_HTTPS_PORT, "9443");
+        endpointProperties.put(BoostProperties.ENDPOINT_HTTPS_PORT, https_port);
+        
+        serverConfig.addBootstrapProperties(endpointProperties);
+        
         // Add war configuration is necessary
```
scottkurz: Hmm...not sure about the default.

awisniew90:
Changed the property names to "boost.http.host", "boost.http.port", and "boost.http.securePort"

scottkurz:
Just a guess, but maybe the change to using a non-root user (see [here](https://hub.docker.com/_/open-liberty)) is responsible for the failure in the docker builds.

PR #290, opened 2019-03-29
Tomee runtime POC

scottkurz:


chyt:
These changes (and more) are already in the `OpenLiberty:tomee` branch. I'm still working on refactoring the runtime detection code, but those changes are in we can just merge from that branch. So I think we are good to close this PR.

PR #300, opened 2019-04-15
Change JPA booster to include JDBC booster as compile scope dep

awisniew90:


awisniew90:
Travis failing due to known issue with travis/mysql port

PR #312, opened 2019-04-29
fix tomee test failures

awisniew90:
1. Added tomee profiles to all IT projects
2. Test code now determined port based on runtime rather than explicitly passing in a port
3. Liberty specific tests (LibertyFeatureVersionIT) are skipped using junit AssumeTrue
4. Spring tests are skipped for Tomee altogether using <excludes> tag in maven failsafe plugin.

Two tests are still failing with what seems like deeper problems that we should open separate issues for:

1. test-dev-release:

Caused by: org.apache.openejb.config.UnknownModuleTypeException: Unknown module type: url=/Users/awisniew/Projects/Liberty-Boost/boost/boost-maven/boost-maven-plugin/target/it/test-dev-release/target/it/release-project/target/apache-tomee/apps/ROOT.jar which can't be a war.

2. test-mpHealth-1.0:

The localhost:<port>/health url throws a 404 on tomee.... not sure why.

### DIFF
--- a/boost-maven/boost-maven-plugin/pom.xml
+++ b/boost-maven/boost-maven-plugin/pom.xml
@@ -97,6 +97,9 @@
 							<streamLogs>true</streamLogs>
 							<properties>
 								<boostRuntime>${boostRuntime}</boostRuntime>
+								<!-- Explicitly set the boost port to 9000 for testing purposes 
+								     across multiple runtimes -->
+								<boost.http.port>9000</boost.http.port>
 							</properties>
 						</configuration>
 						<executions>
--- a/boost-maven/boost-maven-plugin/src/it/test-dev-release/pom.xml
+++ b/boost-maven/boost-maven-plugin/src/it/test-dev-release/pom.xml
@@ -36,6 +36,7 @@
 					</pomIncludes>
 					<properties>
 						<boostRuntime>${boostRuntime}</boostRuntime>
+						<boost.http.port>${boost.http.port}</boost.http.port>
 					</properties>
 				</configuration>
 				<executions>
--- a/boost-maven/boost-maven-plugin/src/it/test-dev-release/src/it/release-project/pom.xml
+++ b/boost-maven/boost-maven-plugin/src/it/test-dev-release/src/it/release-project/pom.xml
@@ -99,7 +99,7 @@
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-failsafe-plugin</artifactId>
-				<version>2.19.1</version>
+				<version>2.22.1</version>
 				<executions>
 					<execution>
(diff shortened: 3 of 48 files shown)


scottkurz:
Also yes, agree the other two failures should be separate issues.   

You might go ahead and open the issues and add an `@Ignore("...issueNum..")` to each.    We haven't had a green build in awhile, so would be nice, since probably these remaining two aren't trivial ones we're going to resolve in a day.

scottkurz:
Another question..just looking at this Travis build: https://travis-ci.com/OpenLiberty/boost/jobs/196420005

The failure doesn't seem to match your description of the two known failures.  Any ideas what's going on?

awisniew90:
@scottkurz - To answer your questions:

1. I think as a model, it would be good to show the port not being hardcoded, although unless we explicitly pass in the port on the command line, the tests wont know what port is being used. We certainly need a discussion around the right way to do this. But as you said, for now, I'll leave the ports hardcoded depending on the runtime defaults. 

2. The "assume" skips the test class altogether. The IT shows up in the maven output like this:

[INFO] Build Summary:
[INFO]   Passed: 1, Failed: 0, Errors: 0, Skipped: 0

But the tests themselves dont run and dont show up in the output

3. Not sure where that constants class should live, since each of these IT projects are treated as separate projects... 

Also - not sure we want to add the @Ignore tag for the failing tests, since they are only failing with tomee. We'd want them to still run with OL

scottkurz:
@awisniew90 - True, `@Ignore` might be going too far.    I think what I'd been thinking (if you could have just read my mind) was another assume with a Rule somehow.    I searched and found [this](https://www.codeaffine.com/2013/11/18/a-junit-rule-to-conditionally-ignore-tests/) example.

scottkurz:
As far as the constants issue... OK, probably not worth at this point adding structure/complexity to solve that problem...just given all else that's on our plate.

chyt:
I think this would be a good point to discuss making the Boost plugin more extensible. I noticed in the test code that we are changing our expected results for the tests based on the selected runtime. I think that this makes the test code less readable and is also difficult to scale with the addition of more runtimes. 

More importantly, I like the idea of being able to implement runtimes independently of the Boost plugin, similar to what is done with the Arquillian container adapaters. Structuring our tests in this way would make it more difficult to move to such an implementation, if we decide to. In that implementation, each runtime implementation would be responsible for its own set of tests. Due to the shared boosters in the Boost scenario, I could imagine us creating a standardized test interface for all the boosters, and having each runtime provide its own implementation with the correct expected values, rather than relying on if/else for runtime detection. 

I think this PR can still go in at this time since we probably need more discussion on the extensibility idea, but I just want to point out now that I see it as a better structure for both the runtime and test implementations.

PR #316, opened 2019-05-03
Create uber jar

awisniew90:
Create runnable jar for Liberty and TomEE regardless of packaging type.


awisniew90:
Replaced by #320

PR #317, opened 2019-05-07
Cdi it

ajm01:
Introduce a new IT test suite for the cdi booster - and also covers jsonP as well? 

This is an open liberty guide specific to cdi and jsonp and is EE based (no microprofile)


boost-maven/boost-boms/booster-ee8-bom/pom.xml
```diff
@@ -60,12 +60,12 @@
...
-				<groupId>io.openliberty.boosters</groupId>
+				<groupId>boost.boosters</groupId>
 				<artifactId>cdi</artifactId>
 				<version>0.2-SNAPSHOT</version>
 			</dependency>
 			<dependency>
-				<groupId>io.openliberty.boosters</groupId>
+				<groupId>boost.boosters</groupId>
 				<artifactId>jsonp</artifactId>
 				<version>0.2-SNAPSHOT</version>
```
scottkurz: 'jsonp' dependency isn't really doing anything here, at least not in Liberty (probably not in TomEE either).   Maybe slightly confusing to leave it in not a huge deal though.

We should start looking at using 'jsonp' together with a JSON provider dependency to use a non-default provider, as a totally separate issue though.

PR #318, opened 2019-05-15
Boost Runtime Adapters

mattbsox:
First pass on the runtime adapter changes for Maven.
- Made the boost.runtimes boosters hold onto the runtime adapters
- Loading Maven project compile dependencies into our plugin
- Checking those dependencies for a boost.runtime jar
- Creating the runtime object from the class that implements RuntimeI in the boost.runtime jar

Also removed the SpringBoot tests and plugin code.


boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/plugin/AbstractMojo.java
```diff
@@ -88,31 +99,34 @@ public void execute() throws MojoExecutionException {
...
+                for(String compilePathElement : project.getCompileClasspathElements()) {
+                    pathUrls.add(new File(compilePathElement).toURI().toURL());
+                }
+
+                URL[] urlsForClassLoader = pathUrls.toArray(new URL[pathUrls.size()]);
+                ClassLoader compileClassLoader = new URLClassLoader(urlsForClassLoader, this.getClass().getClassLoader());
+                List<Class<?>> runtimes = ClassFinderUtil.findClassesImplementing(RuntimeI.class, "boost.runtimes", compileClassLoader, logger);
+
+                if (runtimes.size() == 1) {
+                    Object o = runtimes.get(0).getConstructor(params.getClass()).newInstance(params);
```
awisniew90: Can we just pass in the `parms` object here rather than creating a new instance?

boost-maven/boost-runtimes/runtime-openliberty/src/main/java/boost/runtimes/LibertyRuntime.java
```diff
@@ -121,183 +115,34 @@ public void doPackage() throws BoostException {
     private void packageLiberty(List<AbstractBoosterConfig> boosterConfigs) throws MojoExecutionException {
         createLibertyServer();
 
-        /**
-         * Whether the packaged Liberty Uber JAR will be the project artifact. This
-         * should be the case in Spring Boot scenarios since Spring Boot developers
-         * expect a runnable JAR.
-         */
-        boolean attach;
+         // Dealing with an EE based app
```
awisniew90: We can probably get rid of this comment since we are only handling EE based apps now

boost-maven/boost-runtimes/runtime-openliberty/pom.xml
```diff
@@ -7,5 +7,76 @@
...
+			<artifactId>maven-plugin-api</artifactId>
+			<version>2.0</version>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.maven.plugin-tools</groupId>
+			<artifactId>maven-plugin-annotations</artifactId>
+			<version>3.4</version>
+			<scope>provided</scope>
+		</dependency>
+		<dependency>
```
awisniew90: Since we are adding these dependencies over in the runtime pom, we probably want to remove them from the boost-maven-plugin pom.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/plugin/AbstractMojo.java
mattbsox: Yup, that's what we're doing. We're getting a constructor that takes a `RuntimeParams` object and then calling `newInstance()` on that constructor.

awisniew90: Ah my mistake!  I missed the ")" after getClass() and thought we were calling newInstance on that.

boost-common/src/main/java/io/openliberty/boost/common/utils/ClassFinderUtil.java
```diff
@@ -0,0 +1,115 @@
+package io.openliberty.boost.common.utils;
```
cherylking: Copyright

boost-maven/boost-maven-plugin/src/it/test-dynamic-update/src/test/java/it/EndpointIT.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2018 IBM Corporation and others.
+ * Copyright (c) 2019 IBM Corporation and others.
```
cherylking: You should not change 2018 to 2019. Just add 2019 such as "2018, 2019"

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/plugin/AbstractMojo.java
```diff
@@ -15,9 +15,19 @@
 import static org.twdata.maven.mojoexecutor.MojoExecutor.groupId;
 import static org.twdata.maven.mojoexecutor.MojoExecutor.plugin;
 
+import java.io.File;
```
cherylking: Should update all copyrights for changed files to include 2019..."2018, 2019"

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/runtimes/RuntimeParams.java
```diff
@@ -0,0 +1,75 @@
+package io.openliberty.boost.maven.runtimes;
```
cherylking: copyright

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/plugin/AbstractMojo.java
```diff
@@ -88,31 +99,34 @@ public void execute() throws MojoExecutionException {
...
+            RuntimeParams params = new RuntimeParams(dependencies, getExecutionEnvironment(), project, getLog(), repoSystem, repoSession, remoteRepos, getMavenDependencyPlugin());
+            try {
+                List<URL> pathUrls = new ArrayList<URL>();
+                for(String compilePathElement : project.getCompileClasspathElements()) {
+                    pathUrls.add(new File(compilePathElement).toURI().toURL());
+                }
+
+                URL[] urlsForClassLoader = pathUrls.toArray(new URL[pathUrls.size()]);
+                ClassLoader compileClassLoader = new URLClassLoader(urlsForClassLoader, this.getClass().getClassLoader());
+                List<Class<?>> runtimes = ClassFinderUtil.findClassesImplementing(RuntimeI.class, "boost.runtimes", compileClassLoader, logger);
```
scottkurz: Hmm... I haven't thought this through extenisvely, but would've expected maybe something like a ServiceLoader extension...e.g. an `src/main/resources/META-INF/services/io.microprofile.boost.RuntimeI` with the contents equal to an impl class name.      It seems this assumes a convention that all impls use the **boost.runtimes** package (or package prefix maybe?  didn't digest code fully).   Is there a precedent for that sort of extension mechanism over something like **ServiceLoader**?

mattbsox: Ya, didn't know about this approach. I'll take a look.

scottkurz: The path would match the intf qualified class name, so for the code as-is it would actually be `src/main/resources/META-INF/services/io.openliberty.boost.common.runtimes.RuntimeI`.

boost-common/src/main/java/io/openliberty/boost/common/boosters/MPHealthBoosterConfig.java
```diff
@@ -20,40 +20,23 @@
...
-            deps.add("org.apache.geronimo:geronimo-health:1.0.1");
-            deps.add("org.apache.geronimo:geronimo-health-common:1.0.1");
-            deps.add("org.eclipse.microprofile.health:microprofile-health-api:1.0");
-        }
+        // if(runtime instanceof TomeeRuntimeI) {
+        //     deps.add("org.apache.geronimo:geronimo-health:1.0.1");
+        //     deps.add("org.apache.geronimo:geronimo-health-common:1.0.1");
+        //     deps.add("org.eclipse.microprofile.health:microprofile-health-api:1.0");
+        // }
         return deps;
```
scottkurz: Will look for the follow-up commit including TomEE to see where this goes.   Understand it's not there now.

boost-maven/boost-maven-plugin/src/it/test-dynamic-update/src/test/java/it/EndpointIT.java
```diff
@@ -75,61 +75,26 @@ public void testServlet() throws Exception {
...
             assertTrue("Unexpected response body",
                     response.contains("Hello World From Your Friends at Liberty Boost EE!"));
-        } finally {
-            method.releaseConnection();
-        }
-    }
-
-    @Test
-    public void testServletAfterCompile() throws Exception {
-        HttpClient client = new HttpClient();
```
scottkurz: Hmm... just curious why we only are removing now.  I can see why we'd not include this for TomEE if it depends on "loose app"... maybe this test hadn't been running the best couple weeks?

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/plugin/AbstractMojo.java
```diff
@@ -88,31 +99,34 @@ public void execute() throws MojoExecutionException {
...
-                logger.info("Detected WebSphere Liberty as target Boost runtime");
-                String runtimeGroupId = "com.ibm.websphere.appserver.runtime";
-                String runtimeArtifactId = "wlp-javaee7";
-                String runtimeVersion = "19.0.0.3";
-                runtime = new LibertyRuntime(dependencies, getExecutionEnvironment(), project, getLog(), repoSystem, repoSession, remoteRepos, getMavenDependencyPlugin(), runtimeGroupId, runtimeArtifactId, runtimeVersion);
-            }
-            else {
-                throw new MojoExecutionException("No target Boost runtime was detected. Please add a runtime and restart the build.");
+            RuntimeParams params = new RuntimeParams(dependencies, getExecutionEnvironment(), project, getLog(), repoSystem, repoSession, remoteRepos, getMavenDependencyPlugin());
+            try {
```
scottkurz: This RuntimeParams looks too complicated.

What I think we do is:  instead of the implementation using **ServiceLoader** to plug-in a Runtime impl, they plug-in a RuntimeFactory impl... so we have two interfaces, RuntimeFactoryI and RuntimeI.     The factory interface is the signature implied by:    `RuntimeI  createRuntime(dependencies, getExecutionEnvironment(), project, getLog())`   (I know that's not my lazy shorthand for listing the actual types).      This assumes that we only need the other stuff:  `repoSystem, repoSession, remoteRepos, ` to get the dependencies?   The maven dependency plugin is something an impl can get itself.

Would that work or am I missing something?

boost-maven/boost-boms/booster-runtimes-bom/pom.xml
```diff
@@ -28,20 +28,20 @@
 				<groupId>boost.runtimes</groupId>
 				<artifactId>openliberty</artifactId>
 				<version>0.1-SNAPSHOT</version>
-				<scope>provided</scope>
+				<scope>compile</scope>
```
scottkurz: Just curious why this change?

boost-common/src/main/java/io/openliberty/boost/common/boosters/CDIBoosterConfig.java
```diff
@@ -25,25 +25,8 @@
 @BoosterCoordinates(AbstractBoosterConfig.BOOSTERS_GROUP_ID + ":cdi")
 public class CDIBoosterConfig extends AbstractBoosterConfig {
 
-    String libertyFeature = null;
-
     public CDIBoosterConfig(Map<String, String> dependencies, BoostLoggerI logger) throws BoostException {
```
scottkurz: Understand if we're treating this as a "work in progress", not wanting to tear away the reflection calls using this.   Otherwise the **AbstractBoosterConfig** approach obviously no longer seems useful.

boost-maven/boost-runtimes/runtime-openliberty/src/it/java/boost/runtimes/utils/BoosterUtil.java
```diff
@@ -0,0 +1,22 @@
...
+            throws BoostException {
+        Map<String, String> map = new HashMap<String, String>();
+        map.put(AbstractBoosterConfig.getCoordinates(booster), version);
+        return map;
+    }
+
+    public static Map<String, String> getJDBCDependency() throws BoostException {
+        return BoosterUtil.createDependenciesWithBoosterAndVersion(JDBCBoosterConfig.class, "0.1-SNAPSHOT");
+    }
+}
```
scottkurz: Understand not wanting to pull out the UTs, but I wouldn't leave this layer for more than a short period...it's confusing and not doing anything useful.   If BoosterConfig(s) are common but these UTs are trying to do something Liberty-specific, let's not confusing things with this extra test layer.    

Until we decide what to do w/ BoosterConfig(s) we can leave it in I guess though.

boost-maven/boost-maven-plugin/src/main/java/io/openliberty/boost/maven/plugin/AbstractMojo.java
scottkurz: Actually, let me amend that.    A factory interface is better than a constructor with **MyImpl(RuntimeParms)**.. but we could collapose this into one interface where **RuntimeI**   does `doPackage(dependencies, getExecutionEnvironment(), project, getLog())`.   I don't think there's any need to try to store "state" for start/stop, etc... so that'd be better than having two interfaces to implement (RuntimeI & RuntimeFactoryI).

PR #320, opened 2019-05-28
Create uber jar

awisniew90:


awisniew90:
@scottkurz - hmm. so we are just creating the uber jar at this point. Not making any decisions on what actually gets installed. i.e. with a packaging type of "war", we create the war file and the uber jar, but the war will be installed. Packaging type of jar doesnt work at this point for other reasons, so when we tackle that, we will need to start the discussion of what actually gets installed.

PR #321, opened 2019-05-31
Refactored OL and TomEE booster configurators

chyt:


boost-maven/boost-runtimes/runtime-openliberty/src/main/java/boost/runtimes/boosters/LibertyBoosterI.java
```diff
@@ -0,0 +1,20 @@
...
+ *******************************************************************************/
+package boost.runtimes.boosters;
+
+import boost.runtimes.LibertyServerConfigGenerator;
+import io.openliberty.boost.common.BoostException;
+
+public interface LibertyBoosterI {
+
+    public String getFeature();
+    public void addServerConfig(LibertyServerConfigGenerator libertyServerConfigGenerator) throws BoostException;
```
awisniew90: I think we should have the `addServerConfig()` method be a part of main set of boosters, not just the LibertyBoosterI implementations. Looking at the JDBC booster for example, every runtime is going to need to add server config for this. If we have this just as part of the main JDBC booster, we can then bring back the ServerConfigGenerator interface and pass that to the `addServerConfig()` method. That method can then just call `ServerConfigGenerator.addDatasource(...)` and execute whatever implementation of that method the runtime has. That would help structure the config generators across runtimes and keep any actual config logic out of the booster implementations.

PR #328, opened 2019-08-01
Pull new boosters

ajm01:
this is the PR currently under review on the microprofile sandbox here: 
https://github.com/scottkurz/microprofile-sandbox/pull/9

there were comments made there and addressed which i pulled over to this repo. 

Any additional comments can be made here.


boost-maven/boost-boms/booster-ee7-bom/pom.xml
```diff
@@ -43,7 +43,7 @@
 				<!-- Booster version distinct from the BOM, boost-maven-plugin version -->
 				<!-- Could also use 7.0, 8.0 for EE7, 8 -->
 				<!-- Probably will be 1.0 for initial GA -->
-				<version>0.1-SNAPSHOT</version>
+				<version>2.0-M1-SNAPSHOT</version>
```
chyt: Conventionally, I believe the naming should be like `2.0.M1-SNAPSHOT`, not `2.0-M1-SNAPSHOT`. That way the final released version looks like `2.0.M1`. At least that is how I have seen it in the past, and how I versioned it in the LMP 3.0.M1 release. Maybe @scottkurz can confirm if we want to proceed with this change across the board?

boost-maven/boost-runtimes/runtime-openliberty/src/test/java/io/openliberty/boost/runtimes/boosters/CDIBoosterTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+public class CDIBoosterTest {
+
+    @Rule
+    public TemporaryFolder outputDir = new TemporaryFolder();
+
+    @Rule
+    public final RestoreSystemProperties restoreSystemProperties = new RestoreSystemProperties();
+
+    // private Map<String, String> getCDIDependency(String version) throws
```
chyt: Delete commented code here

boost-maven/boost-boms/booster-ee7-bom/pom.xml
scottkurz: I don't have a strong opinion, and in reading 
https://maven.apache.org/pom.html#Version_Order_Specification
it seems we have a lot of flexibility and yet can still end up with a well-defined ordering scheme.

Maybe it depends what we are tracking exactly, JAX-RS API versions vs. Liberty feature versions.   If we say we're tracking Java/Jakarta EE versions which might have an **x.y.z** version name then I could see using (-) to indicate where the version transitions from "theirs" to "ours"...  so more like Andy's existing `2.0-M1-SNAPSHOT`. 

But if there were other conventions maybe that'd be helpful.  

Also I know we threw out the "M1" idea.. but it might be more suggestive to keep this as "alpha" for now... so 2.0-a1-SNAPSHOT or 2.0-alpha-1-SNAPSHOT .    

What do you think?

scottkurz: Hmm..so for e.g., CDI 2.0 the latest commit uses:
`2.0.0.M1-SNAPSHOT`
Charles' suggestion was:  `2.0.M1-SNAPSHOT` and mine was `2.0-a1-SNAPSHOT`.  

I think either is fine... but I don't want to add an extra .0, like `2.0.0.... ` . I haven't digested every point in https://maven.apache.org/pom.html#Version_Order_Specification  but my reading is that as long as we're consistent going forwards, either would work. 

We could figure out the M1 vs a1 later than this PR.

scottkurz: BTW, looks like the Travis failure https://api.travis-ci.org/v3/job/567141295/log.txt shows you need to still change the version in the boosters and the IT poms.

scottkurz:
As we expand out the matrix, wondering how to expand on the table in https://github.com/dev-tools-for-enterprise-java/boost/wiki/Current-Status-of-Boost-Technology-Boosters  to give a better sense of what we've done and what doc we would want to write.

Some questions we might answer to go beyond a "yes/no":
* What specific config do we generate for each runtime?
* What specific Boost properties may be used?
* Any special testing notes?    If we had a baseline expectation of, say, Open Liberty + TomEE integration + Open Liberty unit test then we could just document the exceptions.

Another approach could be to have an asciidoc page checked in as source for each booster, which could be generated into html later.

boost-maven/boost-boms/boost-ee7-apis-bom/pom.xml
```diff
@@ -36,6 +36,16 @@
                 <artifactId>javax.transaction-api</artifactId>
                 <version>1.2</version>
             </dependency>
+            <dependency>
```
scottkurz: This PR, hitting this many files, is probably a good time to try to format our XML more consistently.   Can you suggest a format, and  apply consistently for the changed parts?   I'd suggest four spaces..that way we can do a simple grep for tab as a check at some point.

boost-common/src/main/java/boost/common/config/ConfigConstants.java
```diff
@@ -100,12 +100,19 @@
     public static final String JDBC_43 = "jdbc-4.3";
     public static final String JPA_21 = "jpa-2.1";
     public static final String JPA_22 = "jpa-2.2";
+    public static final String CDI_12 = "cdi-1.2";
     public static final String CDI_20 = "cdi-2.0";
     public static final String MPHEALTH_10 = "mpHealth-1.0";
+    public static final String MPOPENAPI_10 = "mpOpenAPI-1.0";
     public static final String MPRESTCLIENT_11 = "mpRestClient-1.1";
+    public static final String MPMETRICS_11 = "mpMetrics-1.1";
```
scottkurz: Not your change at fault but it'd be nice if the constant reflected the fact that it was a Liberty feature.

boost-maven/boost-boms/boost-ee8-apis-bom/pom.xml
```diff
@@ -46,6 +46,11 @@
 				<artifactId>javax.json-api</artifactId>
 				<version>1.1</version>
 			</dependency>
+			<dependency>
+				<groupId>javax.json.bind</groupId>
+				<artifactId>javax.json.bind-api</artifactId>
+				<version>1.0</version>
+			</dependency>
 		</dependencies>
```
scottkurz: Should we start by pulling in the EE 8 aggregate dependency (same thought for EE 7 and MPs)?   We could always override I believe.

boost-maven/boost-runtimes/runtime-openliberty/src/test/java/io/openliberty/boost/runtimes/boosters/MPFaultToleranceBoosterTest.java
```diff
@@ -0,0 +1,64 @@
...
+     * 
+     */
+    @Test
+    public void testMPFaultToleranceBoosterFeature11() throws Exception {
+
+        LibertyServerConfigGenerator serverConfig = new LibertyServerConfigGenerator(
+                outputDir.getRoot().getAbsolutePath(), logger);
+
+        LibertyMPFaultToleranceBoosterConfig libMPFTConfig = new LibertyMPFaultToleranceBoosterConfig(BoosterUtil.createDependenciesWithBoosterAndVersion(LibertyMPFaultToleranceBoosterConfig.class, "1.1.0.M1-SNAPSHOT"), logger);
+
```
scottkurz: Seems less than ideal that we'll need to change this to toggle btw. release and snapshot.   This should be a variable like "RELEASE_QUALIFIER" or something.

boost-maven/boost-runtimes/runtime-openliberty/src/test/java/io/openliberty/boost/runtimes/utils/BoosterUtil.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+import boost.common.BoostException;
+import boost.common.boosters.AbstractBoosterConfig;
+import boost.common.boosters.JDBCBoosterConfig;
+
+public class BoosterUtil {
+
```
scottkurz: This needs some Javadoc..could be done later.

boost-common/src/main/java/boost/common/config/ConfigConstants.java
awisniew90: Given that this class keeps growing, it would be good to break this up. Perhaps having a "LibertyConfigConstants" class that lives in the Liberty runtime adapter or defining the constants in a more suitable class rather than a constants class. Maybe in the LibertyServerConfigGenerator class?  Or define them more granularly in the Liberty booster config classes themselves?

awisniew90: > Not your change at fault but it'd be nice if the constant reflected the fact that it was a Liberty feature.

Given that this class keeps growing, it would be good to break this up. Perhaps having a "LibertyConfigConstants" class that lives in the Liberty runtime adapter or defining the constants in a more suitable class rather than a constants class. Maybe in the LibertyServerConfigGenerator class or define them more granularly in the Liberty booster config classes themselves. Either way, I think they should be pulled out of common since they are runtime specific. I can open an issue for this.

boost-maven/boost-boms/boost-ee8-apis-bom/pom.xml
awisniew90: @scottkurz  - do you mean this?

<dependency>
    <groupId>javax</groupId>
    <artifactId>javaee-api</artifactId>
    <version>8.0</version>
</dependency>

boost-maven/boost-boms/booster-ee7-bom/pom.xml
```diff
@@ -1,64 +1,76 @@
...
+                <type>pom</type>
+            </dependency>
+            <dependency>
+                <groupId>boost.boosters</groupId>
+                <artifactId>jaxrs</artifactId>
+                <!-- Booster version distinct from the BOM, boost-maven-plugin 
+                    version -->
+                <!-- Could also use 7.0, 8.0 for EE7, 8 -->
+                <!-- Probably will be 1.0 for initial GA -->
+                <version>2.0-0-M1-SNAPSHOT</version>
```
scottkurz: This looks a bit less readable than I thought...can we change this from `2.0-0-M1-SNAPSHOT` to `2.0-1.0-M1-SNAPSHOT`... so to start we'll align with the plugin version (which, pre-1.0, will be `1.0-M1-SNAPSHOT`  - though of course may diverge over time).

scottkurz: Of course I meant to make similar changes for all boosters.

scottkurz: Moving this comment to PR #337 and counting this "resolved".

PR #355, opened 2019-08-21
Fix logger to output debug messages at the debug level.

### LINKED ISSUE #333: Noisy debug logging by default in dependency detection
If this is debug level logging why is it on by default:

> [debug] Found dependency while processing project: boost.boosters:jaxrs:0.2-SNAPSHOT:jar:provided

scottkurz:

### DIFF
--- a/boost-maven/boost-maven-plugin/src/main/java/boost/maven/plugin/AbstractMojo.java
+++ b/boost-maven/boost-maven-plugin/src/main/java/boost/maven/plugin/AbstractMojo.java
@@ -29,6 +29,7 @@ import org.apache.maven.execution.MavenSession;
 import org.apache.maven.model.Plugin;
 import org.apache.maven.plugin.BuildPluginManager;
 import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.logging.Log;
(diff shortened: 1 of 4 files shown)


scottkurz:
@chyt and @mattbsox  please see what you think about this.

They key addition is to leverage the MojoSupport getLog() by doing an init() first and wrappering it with BoostLogger.....BUT.....though I see you already made BoostLogger a MojoSupport subclass...that didn't seem to work.  It didn't seem to have the right context to use it without initializing, or to initialize it (given that it's not actually a Mojo).  

So I instead have BoostLogger wrappering the MojoSupport Log

        	init();
        	BoostLogger boostLogger = new BoostLogger(getLog());


I didn't really know what to do with the path in which the Runtime impls need to create a BoostLogger... so I just have them wrapping BoostLogger creating the SystemStreamLogger (which is I think what the code was doing before).   

Not sure if that would lead to a similar problem (with the DEBUG coming out even without -X).

But in any case, this at least works for the Maven path... DEBUG logged if and only if -X is present.

Matt, hopefully this doesn't intersect with your current factoring too badly.

scottkurz:
Overriding the need for review since everyone is busy, and since I think this clutters up the log making everyday debugging more difficult.

PR #362, opened 2019-09-05
Transition to new groupid

ajm01:
updated to the new groupid org.microshed and changed LMP versions


scottkurz:
We should do a package rename to match the groupid change ..not that it's necessarily an external.

PR #370, opened 2019-09-10
Update liberty runtime version to 19008

scottkurz:


scottkurz:
This works against the latest in ci.maven (added that to the Travis build instead of relying on latest SNAPSHOT released).

PR #376, opened 2019-09-12
mp30 support

awisniew90:
Microprofile 3.0 support for Open Liberty.  TomEE does not yet support 3.0. Will need to enable that support here in the future.


scottkurz:
Superseded by #380

PR #380, opened 2019-09-13
Adam mp30

scottkurz:


scottkurz:
An equivalent branch passed Travis (I've since rebased)

PR #381, opened 2019-09-13
move boost var names to use underscores vs dots

ajm01:
switch the boost property names from using a '.' separator to using a '_' char

### DIFF
--- a/boost-common/src/main/java/boost/common/config/BoostProperties.java
+++ b/boost-common/src/main/java/boost/common/config/BoostProperties.java
@@ -39,7 +39,7 @@ public final class BoostProperties {
 
     public static final String AES_ENCRYPTION_KEY = "boost_aes_key";
 
-    public static final String INTERNAL_COMPILER_TARGET = "boost_internal_compiler_target";
+    public static final String INTERNAL_COMPILER_TARGET = "boost.internal.compiler.target";
 
     /**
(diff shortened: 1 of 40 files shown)


boost-common/src/main/java/boost/common/config/BoostProperties.java
```diff
@@ -19,24 +19,27 @@
...
+    public static final String DATASOURCE_USER = "boost_db_user";
+    public static final String DATASOURCE_PASSWORD = "boost_db_password";
+    public static final String DATASOURCE_CREATE_DATABASE = "boost_db_createDatabase";
+    public static final String DATASOURCE_URL = "boost_db_url";
 
-    public static final String AES_ENCRYPTION_KEY = "boost.aes.key";
+    public static final String AES_ENCRYPTION_KEY = "boost_aes_key";
 
-    public static final String INTERNAL_COMPILER_TARGET = "boost.internal.compiler.target";
+    public static final String INTERNAL_COMPILER_TARGET = "boost_internal_compiler_target";
```
scottkurz: It might be slightly better to leave this one alone.. since it's not an external,not part of the app+server config.

ajm01: ok, reverting that one back to dots

PR #384, opened 2019-09-13
Snapshot 0.2

scottkurz:


boost-common/src/main/java/boost/common/boosters/AbstractBoosterConfig.java
```diff
@@ -36,32 +36,32 @@
...
+    public static final String MP_FAULTTOLERANCE_VERSION_11 = "1.1" + boostVersion + snapShot;
+    public static final String MP_FAULTTOLERANCE_VERSION_20 = "2.0" + boostVersion + snapShot;
+    public static final String MP_OPENAPI_VERSION_10 = "1.0" + boostVersion + snapShot;
+    public static final String MP_OPENAPI_VERSION_11 = "1.1" + boostVersion + snapShot;
+    public static final String MP_OPENTRACING_VERSION_11 = "1.1" + boostVersion + snapShot;
+    public static final String MP_OPENTRACING_VERSION_12 = "1.2" + boostVersion + snapShot;
+    public static final String MP_OPENTRACING_VERSION_13 = "1.3" + boostVersion + snapShot;
+    public static final String MP_RESTCLIENT_VERSION_11 = "1.1" + boostVersion + snapShot;
+    public static final String MP_RESTCLIENT_VERSION_12 = "1.2" + boostVersion + snapShot;
+    public static final String MP_RESTCLIENT_VERSION_13 = "1.3" + boostVersion + snapShot;
```
chyt: Looking for these versions will break once the release is made...I think the solution is to remove `-SNAPSHOT` from these versions and then match using `startsWith` or `contains` instead of `equals`.

scottkurz:
Superseded by #385

Issue #388, opened 2019-09-16
Suggest not to set appsDirectory=apps when installing app in boost.


patricktiu:
Since the LMP has a "smart" default for the `appsDirectory`, you might not need to set it yourself. 

See the appsDirectory smart default documentation: 
https://github.com/OpenLiberty/ci.maven/blob/master/docs/deploy.md

`boost.runtimes.openliberty.LibertyRuntime`:
```
    /**
     * Invoke the liberty-maven-plugin to run the install-app goal.
     */
    private void installApp(String installAppPackagesVal) throws MojoExecutionException {

        Element deployPackages = element(name("deployPackages"), installAppPackagesVal);
        Element serverNameElement = element(name("serverName"), serverName);

        Xpp3Dom configuration = configuration(deployPackages, serverNameElement, getRuntimeArtifactElement());
        configuration.addChild(element(name("appsDirectory"), "apps").toDom());

        executeMojo(getPlugin(), goal("deploy"), configuration, env);
    }
```

cherylking:
There are other defaults specified as well that do not need to be...like this one:

`                        element(name("outputDirectory"), "target/liberty-alt-output-dir"),`

scottkurz:
@cherylking  We just put that back in last week since @chyt said it wasn't actually defaulted.   Anyway, will save that for later since I have the Travis running in #389.

scottkurz:
OK, I see the change.  Will delete **outputDirectory** too..thx @cherylking

patricktiu:
I can't remember but there is a reason to use a non-default `outputDirectory` for `package` goal.  

Would that be running `package` goal with a running server?

patricktiu:
I think @cherylking is right, the default `outputDirect` for `package` goal is `target/liberty-alt-output-dir` now in 3.0.1 and it is different from the one in `run` goal.  So you don't need it.

PR #389, opened 2019-09-16
Use default appsDirectory parm in deploy goal

scottkurz:


scottkurz:
Fixes #388.

scottkurz:
This doesn't work because LMP looks for the app configured in src server.xml  but Boost is generating it into target server.xml.  https://github.com/OpenLiberty/ci.maven/blob/master/liberty-maven-plugin/src/main/java/io/openliberty/tools/maven/server/PluginConfigSupport.java#L322-L329

Issue #360, opened 2019-09-04
Add ability to specify liberty runtimeVersion


scottkurz:

scottkurz:
This should be done in a way consistent with LMP, e..g see proposed PR https://github.com/OpenLiberty/ci.maven/pull/547/files, where the property **libertyRuntimeVersion** is proposed.

scottkurz:
Fixed by #397

PR #397, opened 2019-09-19
Issue 360 add support for specifying liberty runtime version

uberskigeek:


boost-maven/boost-runtimes/runtime-openliberty/src/main/java/org/microshed/boost/runtimes/openliberty/LibertyRuntime.java
```diff
@@ -79,6 +81,7 @@ public LibertyRuntime(RuntimeParams runtimeParams) {
         this.projectBuildDir = project.getBuild().getDirectory();
         this.libertyServerPath = projectBuildDir + "/liberty/wlp/usr/servers/" + serverName;
         this.mavenDepPlugin = runtimeParams.getMavenDepPlugin();
+        this.runtimeVersion = boostProperties.getProperty("boost_runtime_version", defaultRuntimeVersion);
```
scottkurz: I'd lean towards using the same property as LMP:    -DlibertyRuntimeVersion.    For one..one less property name in the ecosystem.   Also... this isn't going to go into server config

boost-maven/boost-runtimes/runtime-openliberty/src/main/java/org/microshed/boost/runtimes/openliberty/LibertyRuntime.java
```diff
@@ -69,6 +70,7 @@ public LibertyRuntime() {
         this.projectBuildDir = null;
         this.libertyServerPath = null;
         this.mavenDepPlugin = null;
+        this.runtimeVersion = defaultRuntimeVersion;
```
scottkurz: Or should we use the RELEASE qualifier instead?   Especially now that there'd be a way to stick to a given, fixed runtime version

boost-maven/boost-maven-plugin/src/it/test-jaxrs-2.0/src/test/java/it/LibertyFeatureVersionIT.java
```diff
@@ -23,17 +23,23 @@
...
         org.junit.Assume.assumeTrue("ol".equals(runtime) || "wlp".equals(runtime));
+        runtimeVersion = System.getProperty("boost_runtime_version");
     }
 
     @Test
     public void testLibertyFeatureVersion() throws Exception {
         File targetFile = new File(SERVER_XML);
         assertTrue(targetFile.getCanonicalFile() + "does not exist.", targetFile.exists());
+        if (runtimeVersion == null)
+            runtimeVersion = defaultRunTimeVersion;
```
scottkurz: Hmm...I'm a bit wary about saying this is part of the IT API (that you get a "boost_runtime_version" passed as a sysprop).    If we do this I'm OK.. but let's only do so explicitly...so let's remove the default here so that it blows up if we fail to provide this property... instead of having logic buried in the test here.

boost-maven/boost-maven-plugin/src/it/test-jaxrs-2.0/src/test/java/it/LibertyFeatureVersionIT.java
```diff
@@ -23,17 +23,23 @@
 
     private static final String JAXRS_20_FEATURE = "<feature>jaxrs-2.0</feature>";
     private static String SERVER_XML = "target/liberty/wlp/usr/servers/defaultServer/server.xml";
+    private static String MESSAGES_LOG = "target/liberty/wlp/usr/servers/defaultServer/logs/messages.log";
+    private static String runtimeVersion;
+    private static String defaultRunTimeVersion = "19.0.0.9";
```
scottkurz: remove..see other comment.

boost-maven/boost-maven-plugin/src/it/test-jaxrs-2.0/src/test/java/it/LibertyFeatureVersionIT.java
```diff
@@ -15,19 +15,82 @@
...
+        runtimeVersion = System.getProperty("libertyRuntimeVersion");
+    }
+
+    /*
+     * Go to Maven and figure out what the latest version of liberty runtime is out
+     * there. Use that version for comparing what level should be found in the
+     * messages.log file when the server is run. Only used when no libertyRunTime
+     * version is provided.
+     */
+    private String getLatesLibertyRuntimeVersion() {
```
scottkurz: This is neat.. I didn't know how to do all this.   That said, I wonder if it creates some risk that we might not get the exact same latest version in certain environments (IDK, maybe when you have certain stuff in your local .m2 repo), and there isn't really a whole lot to the  boost layer that we're gaining coverage of here.   Boost would be doing the "right thing" even if we got back say 19.0.0.11  instead of 19.0.0.12.    I think we could lose this assertion.. but if we keep it, I would guard it with an error message that spells out in more detail that if it ever fails, well, it could be because of this range check, which doesn't mean Boost is broken, etc.

boost-maven/boost-maven-plugin/src/it/test-jaxrs-2.0/pom.xml
```diff
@@ -59,6 +59,31 @@
...
+        <dependency>
+            <groupId>org.eclipse.aether</groupId>
+            <artifactId>aether-transport-http</artifactId>
+            <version>1.1.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.maven</groupId>
+            <artifactId>maven-aether-provider</artifactId>
+            <version>3.3.9</version>
+        </dependency>
```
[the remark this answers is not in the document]
uberskigeek: Ah yes!

boost-maven/boost-maven-plugin/src/it/test-jaxrs-2.0/invoker.properties
```diff
@@ -0,0 +1,3 @@
+invoker.goals.1 = clean install -DlibertyRuntimeVersion=19.0.0.7
+invoker.goals.2 = clean install -DlibertyRuntimeVersion=19.0.0.8
+invoker.goals.3 = clean install
```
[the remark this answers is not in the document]
uberskigeek: K

boost-maven/boost-maven-plugin/src/it/test-jaxrs-2.0/src/test/java/it/LibertyFeatureVersionIT.java
```diff
@@ -56,4 +119,37 @@ public void testLibertyFeatureVersion() throws Exception {
...
+        // Check the messages.log file for the
+        // Libberty runtime version.
+        boolean found = false;
+        BufferedReader br = null;
+
+        try {
+            br = new BufferedReader(new FileReader(MESSAGES_LOG));
+            String line;
+            while ((line = br.readLine()) != null) {
+                if (line.contains(runtimeVersion)) {
```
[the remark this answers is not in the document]
uberskigeek: Ah, Cause I didn't know about that.

boost-maven/boost-runtimes/runtime-openliberty/src/main/java/org/microshed/boost/runtimes/openliberty/LibertyRuntime.java
```diff
@@ -79,6 +81,7 @@ public LibertyRuntime(RuntimeParams runtimeParams) {
         this.projectBuildDir = project.getBuild().getDirectory();
         this.libertyServerPath = projectBuildDir + "/liberty/wlp/usr/servers/" + serverName;
         this.mavenDepPlugin = runtimeParams.getMavenDepPlugin();
+        this.runtimeVersion = boostProperties.getProperty("libertyRuntimeVersion", defaultRuntimeVersion);
```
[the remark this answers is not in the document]
uberskigeek: k

boost-maven/boost-maven-plugin/src/it/test-jaxrs-2.0/src/test/java/it/LibertyFeatureVersionIT.java
```diff
@@ -56,4 +122,41 @@ public void testLibertyFeatureVersion() throws Exception {
...
+
+        if (runtimeVersion == null) {
+            // This relies on a scan of the maven repository to find the lastest version of
+            // liberty installed. Not sure what subtle issues this might cause.
+            // if there are issues matching the runtime versions then
+            // this may be the culprit and not actually a Boost issue.
+            runtimeVersion = getLatesLibertyRuntimeVersion();
+        }
+        String assertMsg = "Runtime version is null. This may be an issue with checkig for the latest runtime version from maven and may not be a boost issue";
+        assertNotNull(runtimeVersion, assertMsg);
```
scottkurz: I think you got the **assertNotNull** args backwards, see [here](http://junit.sourceforge.net/javadoc/org/junit/Assert.html#assertNull(java.lang.String,%20java.lang.Object))

boost-maven/boost-maven-plugin/src/it/test-jaxrs-2.0/src/test/java/it/LibertyFeatureVersionIT.java
```diff
@@ -56,4 +122,41 @@ public void testLibertyFeatureVersion() throws Exception {
...
+
+        Properties libertyProps = new Properties();
+        String propsRuntime = null;
+        FileInputStream fis = null;
+        try {
+            fis = new FileInputStream(LIBERTY_PROPERTIES);
+            libertyProps.load(fis);
+            propsRuntime = libertyProps.getProperty(LIBERTY_VERSION);
+            assertNotNull(propsRuntime, "Property runtime empty");
```
scottkurz: assertNotNull args

boost-maven/boost-maven-plugin/src/it/test-jaxrs-2.0/src/test/java/it/LibertyFeatureVersionIT.java
```diff
@@ -56,4 +122,41 @@ public void testLibertyFeatureVersion() throws Exception {
...
+        boolean found = false;
+
+        if (runtimeVersion == null) {
+            // This relies on a scan of the maven repository to find the lastest version of
+            // liberty installed. Not sure what subtle issues this might cause.
+            // if there are issues matching the runtime versions then
+            // this may be the culprit and not actually a Boost issue.
+            runtimeVersion = getLatesLibertyRuntimeVersion();
+        }
+        String assertMsg = "Runtime version is null. This may be an issue with checkig for the latest runtime version from maven and may not be a boost issue";
```
scottkurz: It's not a null I'm worried about... I'm worried below when you compare the version calculated by `getLatesLibertyRuntimeVersion` with the one expected by the test.

scottkurz:
Fixes #360

PR #402, opened 2019-09-27
Add DCO signoff instructions

uberskigeek:
This is to add instructions and tools for signing off on a developer certificate of origin. 
Signed-off-by: Mike Andrasak <andrasak@us.ibm.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,7 @@
+# Contributing
+The boost repository requires DCO signing.
+By adding a signed off line to your commit, you verify the following, [Developer Certificate of Origin](https://developercertificate.org).
+
+You can add the signed off line to your commit by adding the -s flag to the git commit command.
+
+If you'd like to automatically sign off all of your commits you can add a prepare-commit-msg hook to do so. A good example can be found in the [Git prometheus](https://github.com/prometheus/prometheus/wiki/DCO-signing) repository DCO wiki entry at the bottom of the page.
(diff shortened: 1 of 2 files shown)


docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,6 @@
+##Contributing
+All boost repositories require DCO signing.
```
scottkurz: We just have one though... All => "the"

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,6 @@
...
+By adding a signed off line to your commit, you sign off this [Developer Certificate of Origin](https://developercertificate.org).
+You can add the signed off line to your commit by adding the -s flag to your commit or have it
```
scottkurz: Can we just link to this prepare-commit-msg somewhere as an idea instead of owning it?

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,6 @@
+##Contributing
```
scottkurz: Let's surface up a bit to ./CONTRIBUTING.md so it's more obvious from the top-level... AND make sure we have a link to it from the top-level README.md

docs/CONTRIBUTING.md
uberskigeek: ok

docs/CONTRIBUTING.md
uberskigeek: ok

PR #407, opened 2019-10-03
Issue 347 Remove possibly unneeded JAXB, activation dependencies

uberskigeek:
Signed-off-by: Mike Andrasak <andrasak@us.ibm.com>


uberskigeek:
#fixes 347

scottkurz:
So does this imply that the other **javax.activation** and **jaxb** references were actually needed then?

